PR #61, opened 2015-02-25
Topic/swagger2.0

cencarnacion:


bryce-anderson:
Wow @cencarnacion, you've been busy. I'll take a look at this very soon. I had fixed the documentation generation on this repos swagger2.0 branch, but I'm sure I'll like your scala immutable model better.

cencarnacion:
@bryce-anderson Yeah. If look at our fork you'll find that I've pushed more bug fixes

swagger/src/main/scala/org/http4s/rho/swagger/SwaggerModelsBuilder.scala
```diff
@@ -0,0 +1,261 @@
...
+import org.http4s.rho.bits.PathAST._
+import org.http4s.rho.bits.QueryAST.{QueryCapture, QueryRule}
+import org.http4s.rho.bits.ResponseGenerator.EmptyRe
+import org.http4s.rho.bits._
+
+import scala.reflect.runtime.universe.Type
+
+import scalaz._, Scalaz._
+
+class SwaggerModelsBuilder(formats: SwaggerFormats) {
```
bryce-anderson: A lot of methods on there are left public. I assume thats for the tests I've yet to scroll down too. Could they be made package private or some other constricted derivative to avoid the one in a million chance someone decides to use them directly?

core/src/main/scala/org/http4s/rho/Router.scala
```diff
@@ -72,15 +72,8 @@ case class CodecRouter[T <: HList, R](router: Router[T], decoder: EntityDecoder[
   override def decoding[R2 >: R](decoder2: EntityDecoder[R2])(implicit t: TypeTag[R2]): CodecRouter[T, R2] =
     CodecRouter(router, decoder orElse decoder2)
 
-  override val headers: HeaderRule = {
```
bryce-anderson: Is the reason this behavior was changed related to swagger support or a more general need? Its now causing a test in ```rho-core``` to fail.

swagger/src/main/scala/org/http4s/rho/swagger/SwaggerSupport.scala
```diff
@@ -10,38 +9,26 @@ import headers.`Content-Type`
...
-
-    swagger.setInfo(info)
-    swagger
-  }
+
+  def apiInfo: Info = Info(title = "My API", version = "1.0.0")
 
   GET / apiPath |>> { () =>
-    val swaggerJson = Json.mapper().writeValueAsString(swagger)
+    val swaggerJson = Json.mapper().writeValueAsString(swagger.map(_.toJModel).get)
```
bryce-anderson: Instead of a ```.get``` can we have an empty json object?

bryce-anderson:
This is excelent work. 
Besides the few things I've commented on, I think this is ready to merge.

Its a shame that your beautiful case class model gets converted back to the swagger types. Is there a reason that we couldn't generate the JSON ourselves directly? One reason why not might be, and I haven't checked if this is the case, that the case class model doesn't fit the schema very well and it might be a pain in the neck. Another reason not to is that by depending on the official swagger spec, we can get a weak 'compile time check' of our compatibilty through breakage with their model.

swagger/src/main/scala/org/http4s/rho/swagger/SwaggerModelsBuilder.scala
cencarnacion: Agree. Furthermore, we can even make the class SwaggerModelsBuilder private to the package swagger.

core/src/main/scala/org/http4s/rho/Router.scala
cencarnacion: Just swagger support. Maybe I shouldn't do it, but it was creating an unnecessary header.

swagger/src/main/scala/org/http4s/rho/swagger/SwaggerSupport.scala
cencarnacion: Agree.

cencarnacion:
Other than having to add a few more details to the scala models, there is no reason we should keep using the swagger types.

swagger/src/main/scala/org/http4s/rho/swagger/SwaggerSupport.scala
cencarnacion: So, should I apply the changes mentioned above and then make a new pull request again?

cencarnacion:
So, should I apply the changes mentioned above and then make a new pull request again?

bryce-anderson:
Whatever is most convenient for you.

swagger/src/main/scala/org/http4s/rho/swagger/SwaggerModelsBuilder.scala
cencarnacion: Where do you stand on this? The entire class should be private to the swagger package or which methods do you think we should expose?

bryce-anderson: I think the whole class should be made private: I foresee little value
outside of our use.

bryce-anderson:
@cencarnacion, I see you've addressed most of the comments, just let me know when you're finished. I'm excited to get this merged. I believe this is pretty much ready for the master branch, right? It appears there are tests for composite models which was the other concern I believe you brought up.

I'd consider doing a point release after this hits master if you're interested.

Thanks again for the great work.

cencarnacion:
@bryce-anderson There is still an issue that cause a rho-core test to fail as you pointed it above. I don't know how to proceed because I far as I can tell we can remove the failing test, but again, I don't know what is the impact outside rho-swagger.

Regarding the composite models issue, I fixed it.

Yes, it's ready for master.

bryce-anderson:
I'm leaning toward just deleting that test: it was intended for functionality that you deleted.

bryce-anderson:
@cencarnacion, I removed the dead tests and have created a PR to merge it to master. I figure its good practice to let any interested parties take a peak before such a big change.

Thanks again for your work. If youre interested in making future contributions, I'm happy to give you committer access.

cencarnacion:
@bryce-anderson Sure. I'm interested in making further contributions, this is just the beginning. :)

PR #63, opened 2015-02-26
Topic/swagger2.0

bryce-anderson:
This merges the swagger 2.0 support. I'm putting this out as a PR for a short review period in case any body has comments or concerns. By short, I mean a few days at most.


bryce-anderson:
Alright, the motion passes.

Issue #69, opened 2015-04-01
java.util.NoSuchElementException: head of empty list


bryce-anderson:
```
> rho-examples[ERROR] java.util.NoSuchElementException: head of empty list
rho-examples[ERROR] 	at scala.collection.immutable.Nil$.head(List.scala:420)
rho-examples[ERROR] 	at scala.collection.immutable.Nil$.head(List.scala:417)
rho-examples[ERROR] 	at org.http4s.rho.swagger.SwaggerModelsBuilder.mkCollectionProperty$1(SwaggerModelsBuilder.scala:244)
rho-examples[ERROR] 	at org.http4s.rho.swagger.SwaggerModelsBuilder.org$http4s$rho$swagger$SwaggerModelsBuilder$$typeToProp$1(SwaggerModelsBuilder.scala:229)
rho-examples[ERROR] 	at org.http4s.rho.swagger.SwaggerModelsBuilder$$anonfun$mkResponse$1.apply(SwaggerModelsBuilder.scala:256)
rho-examples[ERROR] 	at org.http4s.rho.swagger.SwaggerModelsBuilder$$anonfun$mkResponse$1.apply(SwaggerModelsBuilder.scala:256)
rho-examples[ERROR] 	at scala.Option.map(Option.scala:146)
...
```

PR #70, opened 2015-04-01
Fix issue #69

cencarnacion:


bryce-anderson:
Would you mind if we use PR #71 to solve this? It has other enhancements as well. It could use a review before merging.

Thanks for getting on this so fast, sorry I didn't let you know I was working on it too.

cencarnacion:
No, not at all.

bryce-anderson:
Cool.

PR #73, opened 2015-04-29
Fix problem with yield 405 when no paths are defined

bryce-anderson:
Before this, paths such as `/foo/bar` would cause request to `/foo` to return a 405 Method Not Allowed, which I believe is wrong.

I'll merge this soon barring objections.


rossabaker:
Ex post facto :+1:

PR #75, opened 2015-05-06
Cache the swagger json.

bryce-anderson:
Because the model is done during class initialization we can count on
the swagger model being constant once its passed to the user. Therefor,
we should cache the result so we don't waste resources transforming
things back and forth into text.


rossabaker:
:+1:

PR #78, opened 2015-07-07
Remove ADT member from HeaderAST

bryce-anderson:
It turns out that there was a redundency in the HeaderAST: HeaderCapture and HeaderMapper are basically the same thing where HeaderMapper(key, identity) == HeaderCapture.

HeaderCapture now can take a default should the header not exist, and can return an alternate response should the header not be suitable in some way.

Some functions were also renamed to better reflect what they do.


kryptt:
lgtm

rossabaker:
:+1:

PR #82, opened 2015-07-09
Add the concept of a RequestFilter

bryce-anderson:
A RequestFilter can be used to examine an incoming request before running the action. It returns a Task, so it can be asynchronous in its evaluation.

### DIFF
--- a/core/src/main/scala/org/http4s/rho/bits/ResultMatcher.scala
+++ b/core/src/main/scala/org/http4s/rho/bits/ResultMatcher.scala
@@ -7,10 +7,16 @@ import scala.reflect.runtime.universe.{ Type, WeakTypeTag }
 import scalaz.concurrent.Task
 
 
-trait ResultMatcher[-R] {
-  def encodings: Set[MediaType]
-  def resultInfo: Set[ResultInfo]
+trait ResultMatcher[-R] { self =>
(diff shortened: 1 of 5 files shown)


core/src/main/scala/org/http4s/rho/bits/ResultMatcher.scala
```diff
@@ -7,10 +7,16 @@ import scala.reflect.runtime.universe.{ Type, WeakTypeTag }
 import scalaz.concurrent.Task
 
 
-trait ResultMatcher[-R] {
```
bryce-anderson: I'm thinking that `ResultMatcher` might be better as a case class with `conv` as a lambda.

bryce-anderson:
This is on hold until #83 is addressed. It will benefit from the `Action` type, if it lands.

core/src/main/scala/org/http4s/rho/bits/ResultMatcher.scala
rossabaker: And `encodings` and `resultInfo` as parameters?  Because of the `addInfo`?

bryce-anderson:
I think this is no longer worthy of PR status. If I decided to revisit it I'll reopen with a new PR.

PR #83, opened 2015-07-09
Topic/action

bryce-anderson:
This PR reduces the type parameter complexity significantly. It does so at the cost of needing to use `Action` when generating a runnable route. I believe its possible to achieve the same simplification without needing `Action`, but since its here, maybe its worth looking at.

__Pros of using `Action`__
* Future work can add helpers to `Action` to transform the 'action' side of the route, much like can be done with the 'routing' side.

__Cons of using `Action`__
* Its a keyword that we don't necessarily need.


bryce-anderson:
To see the most important differences, use your browser in page search for term `F]`: these are they type parameters that have largely been removed.

core/src/main/scala/org/http4s/rho/RouteExecutable.scala
```diff
@@ -17,22 +17,18 @@ trait RouteExecutable[T <: HList] extends TypedBuilder[T] {
...
 
-  /** Create a [[RhoAction]] from this [[RouteExecutable]] with the provided converters */
-  def makeAction[F](f: F, hf: HListToFunc[T, F]): RhoAction[T, F]
+//  /** Create a [[RhoAction]] from this [[RouteExecutable]] with the provided converters */
+  def makeAction(action: Action[T]): RhoRoute[T]
 
   /** Compiles a HTTP request definition into an action */
-  final def |>>[F, R](f: F)(implicit hf: HListToFunc[T, F], srvc: CompileService[F, R]): R =
-    srvc.compile(makeAction(f, hf))
+  final def |>>[R](action: Action[T])(implicit srvc: CompileService[R]): R =
```
rossabaker: Dimwitted question: does it defeat the purpose if this is overloaded with something like the old signature to bundle into an `Action` and then call this?

bryce-anderson: The overload doesn't work: the compiler complains about not being able to find `HListToFunc` for the `Action` type. I'm guessing that when scala finds functions of the form `def foo(x: RealType)(implicit ...); def foo(y:Y)(implicit ev: T[Y],...)` then things just head south. We could have a different operator...

bryce-anderson: Maybe there is another change in order that will make this simpler: if we make `HListToFunc` explicitly return an `Action` instead of all the components that make up an action, we can make the trivial `HListToFunc[Action]` and only need one operator.

rossabaker: That would be good. `Action` seems like an internal detail and not something to make prominent in an ideal DSL.

bryce-anderson: I'm not sure I agree that `Action` should be viewed as an implementation detail: with a concrete type like `Action` we can manipulate it: make before filters etc.

bryce-anderson: I can't get the compiler to support both `Action` and the `F` that it came from. Not sure why.

bryce-anderson:
Is there a reason I shouldn't merge this? I have a few other things on deck that need to be rebased onto this then submitted as PRs.

PR #84, opened 2015-07-10
PathTree doesnt return a MethodNotAllowed on OPTIONS requests

bryce-anderson:
@kryptt, a quick review would be appreciated. I believe this is what we discussed on gitter.


kryptt:
I think there a COPY/PASTE typo on the last test. 
Otherwise this looks good to me :+1:

PR #85, opened 2015-07-15
Topic/more path tree

bryce-anderson:
This PR changes the internals of the `PathTree` type and the `ParserResult` (also renamed) type.

The `PathTree` stuff is transparent to the current use but makes the tree structure substantially more functional in nature. These changes make it easer to modify the tree, and perhaps in the future it can be used as the only type of tree for representing path structure.

`ParserResult` was renamed to `ResultResponse` and `ValidationFailure` was removed from the ADT. `ValidationFailure` had little distinction from `ParserFailure` except the type of result that it held. Therefor, `ParserFailure` how holds a `FailureReason` which can be turned into a `Task[Response]` if need be.


bryce-anderson:
Any comments on this one? If all stay quiet, I'll probably merge by the end of the week.

rossabaker:
Nothing jumps out at me as bad, but it's a lot of code to go through.  Is there any area in particular you think is shaky and deserves an extra look?

bryce-anderson:
I think its pretty good too. Nothing in particular worth noting. Its mostly a cleanup.

PR #87, opened 2015-08-01
Fix Prepend type param ordering

### LINKED ISSUE #86: weird parameter combination error
// this compiles
      GET / root / "foo" +? param[Option[String]]("bar1") & param[Option[String]]("bar2") & param[Int]("bar3", -1) |>> { (request: org.http4s.Request, b1: Option[String], b2: Option[String], b3: Int) =>
        Ok("ok")
      }
  
      // this does not
      val params = param[Option[String]]("bar1") and param[Option[String]]("bar2") and param[Int]("bar3", -1)
      GET / root / "foo" +? params |>> { (request: org.http4s.Request, b1: Option[String], b2: Option[String], b3: Int) =>
        Ok("ok")
      }

What's the difference between the two ?

bryce-anderson:
closes #86

The order of the type arguments to a Shapeless Prepend were reversed leading to incorrect type.

### DIFF
--- a/core/src/main/scala/org/http4s/rho/Router.scala
+++ b/core/src/main/scala/org/http4s/rho/Router.scala
@@ -37,7 +37,7 @@ case class Router[T <: HList](method: Method,
 {
   override type HeaderAppendResult[T <: HList] = Router[T]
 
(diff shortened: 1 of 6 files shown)


bryce-anderson:
I'll merge this in the next couple days if there are no objections.

Issue #89, opened 2015-08-05
how does rho generate swagger page for a custom param ?


ShengC:
I have made up a `StringParser` to map a parameter (which is basically a json string) to a case class in one of the routes. Now I wish I can show in  `Swagger` page that one of the parameters for this rest api call should be that class. It seems like all param types that can be properly shown in `Swagger` page is defined as an instance of `rho.swagger.TypeBuilder.DataType` which unfortunately is a `sealed trait`. So the question is is it possible to show a custom param type in `Swagger` page at all ?

ShengC:
Guess this would be more clear with an example

    case class Foo(x: Int, y: String)

    val fooParser = new StringParser[Foo] {
        override val typeTag: Some[TypeTag[Foo]] = Some(implicitly[TypeTag[Foo]])
        override def parse(s: String): ParserResult[Foo] = {
          scalaz.\/.fromTryCatchNonFatal(JsonMethods.parse(s).extract[Foo]) match {
            case scalaz.-\/(t) => ParserFailure(t.getMessage())
            case scalaz.\/-(t) => ParserSuccess(t)
          }
        }
     }    

in rho

    GET / "foo" +? param[Option[Foo]]("foo") |>> { (foo: Option[Foo]) =>  Ok(s"received $foo") }

although swagger.json shows for the `foo` route, it takes one parameter typed as `Foo`, but the swagger page just shows `undefined` type for this param. I think I probably need a custom `SwaggerFormats`. If yes, can you point to where I should start with ?

bryce-anderson:
I _think_ it will be a bit more complex than adding a `SwaggerFormats`. I'm not sure the [`SwaggerModelBuilder`](https://github.com/http4s/rho/blob/master/swagger/src/main/scala/org/http4s/rho/swagger/SwaggerModelsBuilder.scala) is currently designed to look in the query or route for models.

bryce-anderson:
I believe we can get this to work without too much pain: much like the [`collectResultTypes`](https://github.com/http4s/rho/blob/master/swagger/src/main/scala/org/http4s/rho/swagger/SwaggerModelsBuilder.scala#L54) and [`collectCodecTypes`](https://github.com/http4s/rho/blob/master/swagger/src/main/scala/org/http4s/rho/swagger/SwaggerModelsBuilder.scala#L60) methods, we need to extract the types from the query or elsewhere from which models may be hiding.

Unfortunately I won't be able to work on this in the short term.

ShengC:
[...]

PR #91, opened 2015-08-07
to address #89

ShengC:
1. Introduce `ComplexDataType` as a new sub class for `DataType` to
distinguish from the primitive data types supported by `swagger` out of
the box.
2. Modify interface of `QueryParameter` and `AbstractProperty` in model to
introduce the `$ref` field for referencing the complex data type. Note
that `RefParameter` and `RefProperty` coming from the Java library do
not work in this case, since they seem to be redeemed for other
purposes. Unfortunately both swagger wiki and java library are very
vague (if not specified at all) how complex type other than the
primitives should be referenced in the query parameter, what I come up
with sort of falls into the realm of “rule-of-thumb” .
(forgot to mention, I also change how the `array` type should be represented, and I believe this is the right way to do it.)
3. In `SwaggerModelsBuilder`, `makeQueryParams` now is changed so that
it absorbs the new format proposed by 2.
4. In `SwaggerModelsBuilder`, create a method to extract complex types
from queries and include them into `collectDefinitions`
5. Add test cases for the change
6. Add example route

### DIFF
--- a/examples/src/main/resources/swagger-ui/index.html
+++ b/examples/src/main/resources/swagger-ui/index.html
@@ -28,7 +28,7 @@
       if (url && url.length > 1) {
         url = decodeURIComponent(url[1]);
       } else {
-        url = "http://localhost:8080/swagger.json";
+        url = window.location.origin + "/swagger.json";
       }
       window.swaggerUi = new SwaggerUi({
         url: url,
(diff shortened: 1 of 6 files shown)


bryce-anderson:
:+1: Looks good to me.

PR #94, opened 2015-08-16
Upgrade to http4s-0.9.0

bryce-anderson:
The change triggers a lot of changes in the tests and a couple in RhoService.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,8 @@
 language: scala
 scala:
-- 2.11.5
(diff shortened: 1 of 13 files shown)


bryce-anderson:
This is in PR form just in case anybody wants to do a quick review. I'll merge this very soon barring objections.

ShengC:
Only thing I care about is whether it would break the swagger documentation :) If no, then :+1:

Issue #99, opened 2015-08-19
`captureMap` causes `internal server error` when `f` throws exception


ShengC:
The right behavior should be sending back `Bad Request` with the reason. I believe this is also an issue for `existsAnd`. We need to wrap a `scala.Try` or `Scalaz.\/` when evaluating `f`

      case HeaderExists(key, f) => req.headers.get(key) match {
        case Some(h) => f(h).fold[ResultResponse[HList]](SuccessResponse(stack))(f => FailureResponse.result(f))

      case HeaderCapture(key, f, default) => req.headers.get(key) match {
        case Some(h) =>
          f(h).fold(f => FailureResponse.result(f), r => SuccessResponse(r::stack))

bryce-anderson:
:+1:

PR #100, opened 2015-08-20
to address #99

ShengC:
sorry for those unnecessary changes of transforming the arrows. My only computer available for use now at home is a macbook air, on which I can only write code in vim (Eclipse would freeze forever)


bryce-anderson:
Those are annoying. I don't want to start introducing unicode characters into the source so those need to be changed before this is merged.

bryce-anderson:
@ShengC, I've removed the unicode arrows in #101 (and used the helpers to check the Response String). Feel free to merge that if your cool with it.

ShengC:
Thanks! I was going to make that change too when convenient. I will merge
it in when I have a chance.

Issue #102, opened 2015-08-25
swagger model builder craps out in case of a `HEAD` defined in route


ShengC:
https://github.com/http4s/rho/blob/master/swagger/src/main/scala/org/http4s/rho/swagger/SwaggerModelsBuilder.scala#L39

https://github.com/http4s/rho/blob/master/swagger/src/main/scala/org/http4s/rho/swagger/models.scala#L191

`wordnik` for `Path` is also missing the `HEAD` request in methods.  

For one, we should add a `default` to guard the `match` of method, and probably we should extend the `wordnik` `Path` class to add the `HEAD` method, though I am not sure if `swagger` can work with `HEAD` request.

bryce-anderson:
Agreed. It looks like [`swagger-core` does support `HEAD` requests](https://github.com/swagger-api/swagger-core/blob/fe2e545806a1c129ce8405c208088452766fb46e/modules/swagger-models/src/main/java/io/swagger/models/Path.java#L18).

bryce-anderson:
Closed by commit 9e746645f1794f974e550ab1d2112c4a16a33654.

PR #103, opened 2015-08-26
address #102

ShengC:


swagger/src/main/scala/org/http4s/rho/swagger/models.scala
```diff
@@ -196,18 +196,53 @@ object models {
...
+    , head             : Option[Operation] = None
     , parameters       : List[Parameter]   = Nil
     , vendorExtensions : Map[String, Any]  = Map.empty
     ) {
-
+    
+    import com.fasterxml.jackson.annotation.{ JsonPropertyOrder, JsonIgnore }
+    
+    @JsonPropertyOrder(Array("get", "post", "put", "delete", "options", "patch", "head"))
+    class ComplementaryPath extends jm.Path {
```
bryce-anderson: I think we just need to bump our swagger-core dep to 1.5.3 and we get these `head` operations, right? (See [here](https://github.com/swagger-api/swagger-core/blob/v1.5.3/modules/swagger-models/src/main/java/io/swagger/models/Path.java#L54))

ShengC: Ok, I think that makes sense to me. I would have to do it after the hours.
If you think this is a trivial enough change, feel free to just make it on
top of mine.

bryce-anderson:
PR #104 includes this work. If need be we can reopen.

PR #105, opened 2015-09-03
Upgrade to http4s-0.10.0

bryce-anderson:
After this is merged I'll cut a release barring objections.


rossabaker:
You don't like http4s' orElse syntax?

bryce-anderson:
No I don't. I prefer to just use the function form because its clear where it came from and in this case it isn't any more trouble.

rossabaker:
Fair enough. LGTM. :+1:

PR #106, opened 2015-10-03
WIP servicesquashing

bryce-anderson:
This PR attempts (and succeeds) at removing the need to use the cake pattern in order to enhance the route structure. It is accomplished by making route tree construction differed until generating the `service`. In the intermediate, all routes are stored by themselves in a simple collection which can be manipulated by 'middleware' type things (`Seq[RhoRoute] => Seq[RhoRoute]`) such as Swagger support.

Questions:
* Should we be operating on the tree or these sequences of routes? What would Swagger prefer? Obviously it works fine now, but would it be better for middleware things to operate on the tree instead? This would result in the `PathTree` becoming public API.
* Related to above, but the signature of 'middleware' like things feels weird.

I like how this has worked out, and it will likely be the basis of something that will be merged into master in the near future. If you have problems or concerns with it, now is the time to speak up.


bryce-anderson:
I think this is pretty close to ready. If someone wants to review this and/or offer suggestions now is the time.

PR #107, opened 2015-10-03
Drop the type parameter from RhoRoute

bryce-anderson:
`RhoRoute` becomes universally quantified (kind of... as much as you can do that in Scala) over its contents and therefor loses its type parameter in favor of a abstract type. I don't think we lose anything anywhere with this. 
The only downsides I see are that its no longer possible to externally match the things inside it such as the `Router` types: the types are just too much for scalac. That required some things to be made into methods on the base type so matching was no longer required such as `validMedia`, `entityType`, and `act`.


bryce-anderson:
I'm feeling less optimistic about this change. I'm going to let it sit, but it may not make it.

bryce-anderson:
Yup, I've made up my mind. This is not going to make it. I'll leave the branch, but that may go in time too.

PR #108, opened 2015-10-03
Construct paths without manually breaking up the uri sections

bryce-anderson:
Currently the behavior would be to accept the string in construction but it would never be matched because uri segments will be broken up by the '/' chars. This PR does the breaking up for the user.


bryce-anderson:
This will be merged before Monday barring objections.

rossabaker:
:+1:

Issue #93, opened 2015-08-10
`HListFunc` only allows function with up to 8 inputs


ShengC:
Can we just change `MatcherHListToFunc` from `private` to `public` ... ? In my use case, I need to pass in more than 8 inputs for the route, and have to extend `HListFunc` to add an implicit to deal with that. Since `MatcherHListToFunc` is `private`, I have to copy that over, which is kind of awkward.

ShengC:
As a side note - why not just use `pattern match` in `HListFunc` to extract the inputs? 

http://stackoverflow.com/questions/18468606/extractor-for-a-shapeless-hlist-that-mimics-parser-concatenation

bryce-anderson:
Sure, changing `MatcherHListToFunc` from `private` to `public` seems OK to me. It is an implementation detail, so we should put a 'not necessarily stable' warning on it.
If you don't mind me asking, how many inputs do you have? That section of code could probably be generated by the build, but that might be more effort than its worth. Additionally, if you create deeper matchers I think they would be welcome here.

There is also a chance there are some functions in shapeless that could do this grunt work for us: https://github.com/milessabin/shapeless/blob/master/core/src/main/scala/shapeless/ops/functions.scala#L29.

From the link your provided, it would seem that using a pattern match will go through a lot of temporary objects which I would like to avoid, though I've not given it a thorough read. There is a good chance the ops functions I've linked to would suffer from the same problem.

ShengC:
I have one route takes 9 inputs, and another takes 10 inputs. Since input can be in the form of `query`, `header` or `path`, I would not be surprised if other folks run into the same issue. 

There is a chance proposal from that link might already be part of `shapeless` given it was 2 years ago, but I might be wrong. Memory footprints might be a problem, but besides that, I don't have a idea of how to make it better.

ShengC:
I've been following Miles' talk and `shapeless` for a while, and realize there actually might be a way to scrap away these boilerplates with something like this,

    import shapeless._
    import shapeless.ops.function._
    import shapeless.syntax.std.function._  

    implicit def instance[T <: HList, F, R](implicit m: ResultMatcher[R], fp: FnToProduct.Aux[F, T => R]): HListToFunc[T, F] = new MatcherHListToFunc[T, F] {
      override def matcher = m
      override def conv(f: F) = (req: Request, hh: T) => {
        m.conv(req, f.toProduct(hh))
      }
    }
    
[...]

PR #110, opened 2015-10-20
address #93

ShengC:
so we have reduced the implicit instance of `HListToFunc` down to just three.

1. `const` addresses the case when the input is a value instead of a function
2. `instance` addresses the case when the input is a function that takes an heterogeneous list as input
3. `instance0` addresses the case when the input is a function that takes an heterogeneous list plus a type of `Request`

implementation relies on the following 3,
1. `FnToProduct.Aux` that generically converts inputs of a function to an heterogeneous list
2. `Reverse.Aux` that generically converts an heterogeneous list into its reverse on type level (Thanks to Travis Brown answering this question in SO: http://stackoverflow.com/questions/33219315/how-to-make-function-taking-reverse-of-an-hlist)
3. Use an implicit `Lazy[ResultMatcher[R]]` to defer the implicit resolution of it. This is important, otherwise compiler will erroneously conclude the implicit does not exist in some scenarios (well I have to say in most of the scenarios).

All code and test compile and pass except one in `RhoServiceSpec` that take an heterogeneous list of `List[String]` as input argument, and a function taking a `Seq[String]` as input. I had to change type of the function taking a `List[String]` instead. I guess `HList` cannot handle the contra-variance (or can it?)


bryce-anderson:
I didn't notice the `Reverse` in there before: that will certainly bring some unnecessary junk instances along with it. The lack of contra-variance is also kind of annoying and might end up breaking a few routes.

Why is `ResultMatcher` now wrapped in a shapeless `Lazy`? It is explicitly unwrapped immediately on execution and thus it only appears to introduce extra junk and hidden locking. Is it necessary to avoid an implicit resolution divergence?

ShengC:
I don't think Reverse will introduce "junk" instance(s). All it does is to
infer type of the reverse of the input HList during compile time. Regarding
use of Lazy, there has to be an proper order of implicit resolution
enforced during compile time, otherwise compiler will make the wrong
conclusion regarding existence of the implicits. You can try get rid of
Lazy wrapper of ResultMatcher, and see what will happen :) Things as such
are just just examples of the limitation of Scala compiler ( I would
strongly recommend checking out Miles' talk in this yr's scaladays. He used
the similar trick to let shapeless derive the type class instances for free)

Regarding the parameter contra variance, that's something I cannot address,
at least with the current knowledge of generic programming. That said, is
it really "that" important ? My gut feeling suggest FP generically speaking
does not like variance. Parboiled 2 and spray routes both apply the similar
technique to feed an HList into a function. AFAIK, neither supports the
parameter contra variance.

bryce-anderson:
The `Reverse` instance will build a reversed version of the parameter `HList`. See the implementation [here](https://github.com/milessabin/shapeless/blob/master/core/src/main/scala/shapeless/ops/hlists.scala#L1659). The reason we need it is because what we really need is a 'type queue' when really `HList` is a 'type stack' so the order of parameters needs to be reversed for the functions in shapeless to be useful. Whether the list reversal has material performance implications is really my question.

I understand that the `Lazy` wrapper can sometimes help with implicit resolution, its easy enough to read the notes in the source [here](https://github.com/milessabin/shapeless/blob/master/core/src/main/scala/shapeless/lazy.scala#L109). My question was whether it was necessary at all or if you could use the `Strict` counterpart defined [here](https://github.com/milessabin/shapeless/blob/master/core/src/main/scala/shapeless/lazy.scala#L148).

I don't think we can get around the variance issue using shapeless. Scala is not a pure FP language and I personally like variance and so I'm happy to use it when it makes sense. That said, its not really important to me here but is a minor annoyance. My concern was more along the lines of 'has someone built something that depends on variance'. I suspect that if they have, it can be easily changed.

ShengC:
inference of type of Reverse of `HList` is done at compile time, so no runtime cost. Regarding whether it can incur runtime cost to invert a material `HList`, this is essentially the same as reverting a `List`, so I think the answer is yes.

I did not try `Strict` wrapper, but when running without the `Lazy` wrapper around `ResultMatcher[R]`, I saw the compiler would go astray trying to resolve the incorrect type associated with the implicits, which eventually came at the wrong conclusion that the implicit could not be found.

Regarding the variance, I think yes if we apply this change, some routes of the project using `rho` will break. But I get the similar feeling this annoyance is ok to live with. If we really want full backwards compatibility, I think we can keep the current implicts in the companion object, and leave these generic implicits as `LowerPriority` to mix in, so the strict type check only happens if they have a function that takes a really long list of parameters. 

Last, I know some projects use the source generation to generate boilerplates like this. But if possible, please let's not go for that option. I like our current code base in that it is well readable, and thus people can contribute at lower cost of their time understanding the code. Things like Macro and Source generation can awe away average users, and thus should be used with discreet.

ShengC:
I have to bump up version of `shapeless` to `2.3.0-SNAPSHOT` to be able to import `shapeless.Strict`. Then compilation failed at `rho-swagger`.

bryce-anderson:
Bummer about the version bump and failure to compile. If we can change
```scala
override def matcher = m.value
// to
override val matcher = m.value // note the 'val' will bring this out of the lazy val
```
then I don't think we will hit the lazy locking but once on initialization.

After that, I think we should merge this with the understanding that if the sky falls or things end up super slow (unlikely) we might have to roll it back.

ShengC:
yes everything compiles and passes the tests with `def` -> `val`, I think difference between `val` and `def` only happens at runtime. If users' feedback strongly demands a pure backwards compatibility or they just cannot live without parameter variance, or it turns out the generic implicits does not live up to our expectation towards it, an alternative to rollback is we can create a `LowerPriority` trait to include these implicits and let the companion object extends from it.

I have to update the PR after hours today, firm proxy won't allow me to push out code.

ShengC:
ok updated PR - need your thumbs up to merge this.

bryce-anderson:
:+1: 
Sorry for dragging this one out. I do like how much boilerplate it removes.

PR #112, opened 2015-10-22
Route construction can add a non-capturing prefix

bryce-anderson:
The idea is that most route construction types can now do the following:
val service = GET / "bar"
val prefixedService = "foo" /: service

We could add capturing prefixes if we wanted to drop support for this from routers but
that would rule out prefixing an already constructed service which is probably the
dominant use case.


ShengC:
I guess I have lots to catch up. Got swamped with `Lucene` and `scodec`, but this looks like a decent enhancement. :+1:

bryce-anderson:
Sold!

PR #113, opened 2015-11-18
[WIP] Topic/applicative

bryce-anderson:
This PR seeks to make the underlying constructs of parsing Applicatives.


bryce-anderson:
I don't think I like this. Its getting axed.

PR #115, opened 2015-11-20
Topic/mergeASTs

bryce-anderson:
The HeaderAST and the QueryAST are very similar in form. (For that matter, so
is the PathAST but we leave that for another day) This commit merges the two
AST's into a RequestAST that contains variants specialized for query and header
capture respectively.

The information needed by swagger et al is really meta data so logically
it makes sense to attach it to the rule tree using a MetaRule. This
results in a dramatic reduction in the cases needed to examine when
moving through the tree, hiding unnecessary information from things like
the actual route evaluation logic.

### DIFF
--- a/core/src/main/scala/org/http4s/rho/ExecutableCompiler.scala
+++ b/core/src/main/scala/org/http4s/rho/ExecutableCompiler.scala
@@ -1,9 +1,7 @@
 package org.http4s
 package rho
 
-import bits.HeaderAST._
-import bits.QueryAST._
-import org.http4s.rho.bits.ResponseGeneratorInstances.BadRequest
+import org.http4s.rho.bits.RequestAST._
 
 import org.http4s.rho.bits._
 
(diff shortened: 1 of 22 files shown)


core/src/main/scala/org/http4s/rho/ExecutableCompiler.scala
```diff
@@ -17,57 +15,16 @@ trait ExecutableCompiler {
...
-
-      case QueryOr(a, b) => runQuery(req, a, stack).orElse(runQuery(req, b, stack))
-
-      case QueryCapture(name, parser, default, _) => parser.collect(name, req.multiParams, default).map(_ :: stack)
-
-      case MetaCons(r, _) => runQuery(req, r, stack)
-
-      case EmptyQuery => SuccessResponse(stack)
-    }
+  def runRequestRules(req: Request, v: RequestRule, stack: HList): ResultResponse[HList] = v match {
```
bryce-anderson: The complexity here was moved to the rho package object.

core/src/main/scala/org/http4s/rho/ExecutableCompiler.scala
```diff
@@ -17,57 +15,16 @@ trait ExecutableCompiler {
...
-    import bits.HeaderAST.MetaCons
-    v match {
-      case HeaderAnd(a, b) => runValidation(req, a, stack).flatMap(runValidation(req, b, _))
 
-      case HeaderOr(a, b) => runValidation(req, a, stack).orElse(runValidation(req, b, stack))
-
-      case HeaderExists(key, f) => req.headers.get(key) match {
-        case Some(h) => 
-          scalaz.\/.fromTryCatchNonFatal( f(h) ) match {
-            case scalaz.-\/(t) => FailureResponse.badRequest(t.getMessage())
```
bryce-anderson: This security leak was changed to a `InternalServiceError` status with a generic message while logging the error. See https://github.com/http4s/rho/pull/115/files#diff-7f2fd9ea04632bd23bdfc7136dcb1b37R156

core/src/main/scala/org/http4s/rho/PathBuilder.scala
```diff
@@ -26,14 +27,12 @@ final class PathBuilder[T <: HList](val method: Method, val path: PathRule)
 {
   type HeaderAppendResult[T <: HList] = Router[T]
 
-  override def headers: HeaderRule = EmptyHeaderRule
-
-  override def query: QueryRule = EmptyQuery
+  override val rules: RequestRule = EmptyRule
```
bryce-anderson: For all the builder types, internally they now use a single set of rules (called `rules`...) instead of two different sets, `headers` and `query`.

core/src/main/scala/org/http4s/rho/bits/RequestAST.scala
```diff
@@ -0,0 +1,19 @@
+package org.http4s
+package rho.bits
+
+object RequestAST {
+
+  sealed trait RequestRule
+
+  case class CaptureRule[T](reader: Request => ResultResponse[T]) extends RequestRule
+
+  case class IgnoreRule(rule: RequestRule) extends RequestRule
```
bryce-anderson: `IgnoreRule` was added so that we can ignore `Unit` types that may pop out of `CaptureRule`s which just validate the request.

core/src/main/scala/org/http4s/rho/bits/UriConverter.scala
```diff
@@ -30,19 +29,19 @@ object UriConverter {
...
-      case QueryOr(a, _) :: rs               => go(a :: rs, acc) // we decided to take the first root
-      case EmptyQuery :: rs                  => go(rs, acc)
+    def go(r: List[RequestRule], acc: Query): Try[Query] = r match {
+      case Nil                                          => Success(acc.reverse)
+      case MetaRule(r, QueryMetaData(n, _, _, _)) :: rs => go(r :: rs, ParamExp(n) :: acc)
+      case MetaRule(r, _) :: rs                         => go(r :: rs, acc)
+      case AndRule(a, b) :: rs                          => go(a :: b :: rs, acc)
+      case (EmptyRule | CaptureRule(_)) :: rs           => go(rs, acc)
+      case IgnoreRule(r) :: rs                          => go(r :: rs, acc)
+      case OrRule(a, _) :: rs                           => Failure(new Exception("Cannot create a query from 'or'ed paths"))
```
bryce-anderson: We no longer decide to arbitrarily pick a path to follow.

bryce-anderson:
This was merged along with a bit of cleanup in commit fe4f4c2128d59bceece5df9409f5913a0351f686.

PR #118, opened 2015-12-15
Add more generic combinators for Header and Query rules

bryce-anderson:
Problem: I wanted to make a HeaderRule that would capture a couple headers and I couldn't without doing things through the untyped channels.

Solution: Make the most generic combinator possible and redefine other combinators in terms of it. That means we now can make rules that directly receive the Request and do whatever it needs to. That doesn't mean its always a good idea and that is reflected in the comments.
Additionally, the Typed rules received new methods that facilitate adding metadata and transforming them into ignore rules which also helps clean up the use of the untyped AST's.


rossabaker:
I defer to those more deeply into Rho, but it passes the sniff test for me.

bryce-anderson:
@rossabaker, thanks for the review. I'll merge this as soon as I find time to work on the project that needs it so it can linger a bit longer in case others have an interest.

PR #120, opened 2016-01-16
Upgrade to http4s 0.12.0

bryce-anderson:
Summary of changes:
* Service and HttpService moved from org.http4s.server to org.http4s so lots of changed imports.
* http4s dropped its home grown DateTime in favor of the java.time.Instant. The swagger TypeBuilder had to be changed accordingly. Also some tests needed changing.


rossabaker:
:+1: 

I will not miss our `DateTime`, even if I miss its `UnixEpoch`.

PR #121, opened 2016-03-25
Upgrade to http4s-0.13-SNAPSHOT

bryce-anderson:
The examples here won't build with 2.11.8. Its fixed by a patch in http4s proper but its not clear if that is a real fix or just a band-aid.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,11 @@ The primary goal of ρ is to provide an easy to use AST with which to build HTTP
 
 Get ρ
 -----
-Rho artifacts are available at Maven Central and snapshots are available from the Sonatype repositories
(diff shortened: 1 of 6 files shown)


rossabaker:
Does this suggest that the build matrix should build against multiple versions of Scala 2.11.x to verify compatibility, even while only one is published?

bryce-anderson:
This appears to be where rho will sit for a while until we can work out the issues with 2.11.8.

rossabaker:
Add a note to the readme and ship it?

bryce-anderson:
Done. And updated the json4s dependencies.

rossabaker:
It occurs to me that rho's JSON use fits into the discussion http4s/http4s#584.  But that's for another ticket.

PR #123, opened 2016-04-28
The DataType of an AnyVal should be the DataType of its underlying type.

etorreborre:
We are using `AnyVals` to model query parameters which are `String` or `Int` and give them a bit more type safety.

I propose that those `AnyVals` are not being considered as independent models but just as wrappers around the underlying value. Not doing this creates an invalid Swagger file for us and I didn't find any convenient work-around.

### DIFF
--- a/swagger/src/main/scala/org/http4s/rho/swagger/TypeBuilder.scala
+++ b/swagger/src/main/scala/org/http4s/rho/swagger/TypeBuilder.scala
@@ -208,6 +208,8 @@ object TypeBuilder {
       } else if (t.isProcess) {
         if (t.typeArgs.nonEmpty) GenArray(fromType(t.typeArgs(1)))
         else GenArray()
+      } else if (klass <:< typeOf[AnyVal]) {
+        fromType(t.members.filter(_.isConstructor).flatMap(_.asMethod.paramLists.flatten).head.typeSignature)
       } else {
         val stt = if (t.isOption) t.typeArgs.head else t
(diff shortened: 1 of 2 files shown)


bryce-anderson:
:+1: This looks good to me. Thanks for the PR! Its great to know more people are using rho!

etorreborre:
I am actually doing an experiment, using rho on top of `akka-http`, just to be able to generate a Swagger file. The work you did on rho is phenomenal and I thought it would be a lot better to try to reuse it instead of starting from scratch.

What is your release cycle with rho? When can I expect a new version with my changes in?

bryce-anderson:
Thanks for the complement! You've been a member of the scala community I have really looked up to so it means a lot to me.

rho development has been relatively slow lately. I've been busy trying to finish graduate school and there are only a few users. However, if you need a release I think its pretty straight forward to cut one.

etorreborre:
>  However, if you need a release I think its pretty straight forward to cut one

If this is easy, it would be nice to release one as soon as #124 is in so that I can remove the duplicated code I have in my project. Then I might do other PRs to address some missing data in the Swagger file, like security info at the beginning of the file (maybe it is already possible to add this info but I didn't find how yet).

bryce-anderson:
rho-0.10.0 should be on its way to Maven. Thanks again for the PR's.

etorreborre:
Got it, thanks a lot for being so reactive!

PR #124, opened 2016-04-28
Extracted a method from SwaggerSupport to create the Swagger spec

etorreborre:
This is useful for us in order to generate a Swagger YAML file which we can commit for review before committing to a given API.

The recursive lazy values in the `apply` method seem to work ok and allow the removal of the `var` :-).

### DIFF
--- a/swagger/src/main/scala/org/http4s/rho/swagger/SwaggerSupport.scala
+++ b/swagger/src/main/scala/org/http4s/rho/swagger/SwaggerSupport.scala
@@ -8,28 +8,42 @@ import headers.`Content-Type`
 import org.http4s.rho.bits.PathAST.TypedPath
 import org.http4s.rho.swagger.models.{Swagger, Info}
 
-import shapeless.HNil
+import shapeless._
 
 object SwaggerSupport {
+
+  /**
+   * Create a RhoMiddleware adding a route to get the Swagger json file
+   * representing the full API
+   */
   def apply(
     swaggerFormats: SwaggerFormats = DefaultSwaggerFormats,
     apiPath: TypedPath[HNil] = "swagger.json",
     apiInfo: Info = Info(title = "My API", version = "1.0.0")): RhoMiddleware = { routes =>
+
+    lazy val swaggerSpec: Swagger =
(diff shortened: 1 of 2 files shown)


swagger/src/main/scala/org/http4s/rho/swagger/SwaggerSupport.scala
```diff
@@ -8,28 +8,40 @@ import headers.`Content-Type`
...
-      lazy val swaggerResponse =
-      Ok(Json.mapper().writeValueAsString(swaggerSpec.toJModel))
-        .putHeaders(`Content-Type`(MediaType.`application/json`))
+    lazy val swaggerSpec: Swagger =
+      createSwagger(swaggerFormats, apiPath, apiInfo)(routes ++ swaggerRoutes)
 
-      "Swagger documentation" ** GET / apiPath |>> { () => swaggerResponse }
-    }.getRoutes()
+    lazy val swaggerRoutes: Seq[RhoRoute[_ <: HList]] = new RhoService {
+      "Swagger documentation" ** GET / apiPath |>> { () => Ok(Json.mapper().writeValueAsString(swaggerSpec.toJModel))
```
bryce-anderson: The one thing I liked about the old behavior is that the swagger spec was memorized. I think that behavior can be preserved.

bryce-anderson:
I like the method extraction: that is a great idea. I do want to keep the serialized json memorized, which I believe should still be possible. That said, I'm open to reasons why that isn't a good idea.

etorreborre:
Isn't it still memoized with a `lazy val`?

etorreborre:
Is there anything I can / should do to fix the build?

etorreborre:
Oh sorry I see what you mean about the serialized json. I'm going to fix that.

etorreborre:
@bryce-anderson please see my last commit.

bryce-anderson:
I've got a PR to try and fix the travis build for PR's testing right now. This PR should actually be a good test for it.
The PR looks good. :+1: Once I get travis happy I'll merge it.

Thanks again!

PR #125, opened 2016-04-28
Try and fix travis PR's

bryce-anderson:
Currently they fail to build because PR's cannot access encryption keys (a good thing).
This PR moves the decryption process to non-pr's that actually need them.


etorreborre:
👍  great!

PR #127, opened 2016-05-07
Cross build for Scalaz 7.2

rossabaker:
Follows the same basic strategy as the main http4s repo.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,6 @@
 language: scala
 scala:
 - 2.11.7
-matrix:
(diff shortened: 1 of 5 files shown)


rossabaker:
This seems to work.  If there are no quick objections, I'll release it as 0.10.1(a) tomorrow.

etorreborre:
That looks good 👍

rossabaker:
0.10.1 and 0.10.1a are on their way to Maven Central.

PR #128, opened 2016-05-20
Return the type string for the NoContent response in swagger

etorreborre:
Otherwise we don't generate a valid Swagger file.


bryce-anderson:
👍 LGTM. I'll merge this tomorrow just in case @rossabaker wanted to look at it.

Thanks for the PR!

PR #129, opened 2016-06-29
[swagger] For a header parameter, set the `required` flag to false if there is a default value

etorreborre:
@bryce-anderson I need something like that to:

 - accept a `X-Flow-Id` header if the client passes on
 - generate a `X-Flow-Id` header otherwise

And have `required=false` in the swagger file for that header parameter.

Does it make sense to you too?

### DIFF
--- a/swagger/src/test/scala/org/http4s/rho/swagger/SwaggerModelsBuilderSpec.scala
+++ b/swagger/src/test/scala/org/http4s/rho/swagger/SwaggerModelsBuilderSpec.scala
@@ -145,6 +145,14 @@ class SwaggerModelsBuilderSpec extends Specification {
         HeaderParameter(`type` = "string", name = "Content-Length".some, description = orStr("Content-MD5"), required = true),
         HeaderParameter(`type` = "string", name = "Content-MD5".some,    description = orStr("Content-Length"), required = true))
     }
+
+    "set required = false if there is a default value for the header" in {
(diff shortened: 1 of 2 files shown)


swagger/src/test/scala/org/http4s/rho/swagger/SwaggerModelsBuilderSpec.scala
```diff
@@ -145,6 +145,14 @@ class SwaggerModelsBuilderSpec extends Specification {
         HeaderParameter(`type` = "string", name = "Content-Length".some, description = orStr("Content-MD5"), required = true),
         HeaderParameter(`type` = "string", name = "Content-MD5".some,    description = orStr("Content-Length"), required = true))
     }
+
+    "set required = false if there is a default value for the header" in {
+      val ra = fooPath >>> captureMapR(`Content-Length`, Option(Ok("5"))) |>> { () => "" }
```
bryce-anderson: I don't think `captureMapR` works like this; its causing travis failure.

bryce-anderson:
With the exception of the bad test, I like it. Once we get that fixed I think its good to go.

etorreborre:
Sorry I had not pushed the right commit. This one should be fine.

etorreborre:
Yes it is! @bryce-anderson

bryce-anderson:
👍 Thanks!

PR #130, opened 2016-06-30
Translate empty maps in a Swagger model to null in the Java model 

etorreborre:
This prevents empty 'examples' and 'headers' fields to be generated for responses.

The biggest contribution of this PR is actually arbitraries for a `Swagger` model. They are far from complete but they allow to test this change.

### DIFF
--- a/swagger/src/main/scala/org/http4s/rho/swagger/models.scala
+++ b/swagger/src/main/scala/org/http4s/rho/swagger/models.scala
@@ -187,6 +187,15 @@ object models {
     , vendorExtensions : Map[String, Any]  = Map.empty
     ) {
 
+    def operations: Seq[Operation] =
+      get.toList ++
+      put.toList ++
+      post.toList ++
+      delete.toList ++
+      patch.toList ++
+      options.toList ++
+      head.toList
+
     def toJModel: jm.Path = {
       val p = new jm.Path
       p.setGet(fromOption(get.map(_.toJModel)))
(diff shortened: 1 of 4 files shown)


swagger/src/main/scala/org/http4s/rho/swagger/models.scala
```diff
@@ -187,6 +187,15 @@ object models {
     , vendorExtensions : Map[String, Any]  = Map.empty
     ) {
 
+    def operations: List[Operation] =
```
bryce-anderson: Not mandatory for merging, and I know this can be a contentious issue, but my personal preference is to return a `Seq` unless its obvious that a `List` poses some unique advantages.

etorreborre: No problem, I can change it. I have taken the habit to be a lot more specific on my collection return types now because I think that it is better to indicate that the collection I am returning will not be infinite for example. But I am fine with changing it.

etorreborre: Done.

bryce-anderson:
👍 Thanks @etorreborre for another good PR. I really appreciate your interest and code. 😄

etorreborre:
@bryce-anderson rebased on master.

bryce-anderson:
On looking again, I noticed a sbt bash script in there. Does it provide value? I haven't put them in my projects but if they are generally useful, I'm not opposed.

etorreborre:
Oh sorry I am generally using it but it is not necessary, I'll remove it.

bryce-anderson:
Thanks @etorreborre! Sorry for all the hassle.

PR #131, opened 2016-07-01
[WIP] Improve error handling during decode

Jacoby6000:
From what I can tell, we currently will ALWAYS output the same failure on a decode error, regardless of what the decoding error was. 

This PR seeks to create a way to generate whatever error message you like. 

Currently, you can retain the old default behavior by importing `org.http4s.rho.bits.EntityParser.defaults._`
### TODO
1. Fix tests
2. Figure out a way to pull out errors so that they may be documented
3. Create helper methods for making generating parsers nice.
4. Whatever comes up in discussion.

### DIFF
--- a/core/src/main/scala/org/http4s/rho/bits/PathTree.scala
+++ b/core/src/main/scala/org/http4s/rho/bits/PathTree.scala
@@ -86,11 +86,11 @@ private[rho] object PathTree {
           }
         }
 
-      case c @ CodecRouter(_, parser) =>
+      case c @ CodecRouter(_, parser, errorHandler) =>
         Leaf { (req, pathstack) =>
           RuleExecutor.runRequestRules(req, c.router.rules, pathstack).map{ i =>
-            parser.decode(req, false).run.flatMap(_.fold(e =>
-              e.toHttpResponse(req.httpVersion),
+            parser.decode(req, false).run.flatMap(_.fold(
+              errorHandler orElse { case e: DecodeFailure => e.toHttpResponse(req.httpVersion)},
               { body =>
                 // `asInstanceOf` to turn the untyped HList to type T
                 route.action.act(req, (body :: i).asInstanceOf[T])
--- a/core/src/main/scala/org/http4s/rho/PathBuilder.scala
+++ b/core/src/main/scala/org/http4s/rho/PathBuilder.scala
@@ -86,7 +86,7 @@ final class PathBuilder[T <: HList](val method: Method, val path: PathRule)
     Router(method, path, h2.rule)
 
   override def decoding[R](decoder: EntityDecoder[R])(implicit t: TypeTag[R]): CodecRouter[T, R] =
-    CodecRouter(>>>(TypedHeader[HNil](EmptyRule)), decoder)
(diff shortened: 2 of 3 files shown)


core/src/main/scala/org/http4s/rho/bits/EntityParser.scala
```diff
@@ -0,0 +1,49 @@
...
+import scala.collection.generic.CanBuildFrom
+import org.http4s.EntityDecoder
+import org.http4s.DecodeFailure
+import scalaz.concurrent.Task
+
+/** Extract a value from the `Request` `Query`
+  *
+  * @tparam A Type of value produced by the parser.
+  */
+ trait EntityParser[A] { self =>
```
bryce-anderson: I don't know this to be true, but I suspect there is a way to do this without adding a wrapper around `EntityDecoder`. I haven't actually explored that, but if it does turn out to be the case, I'd like to avoid a wrapper type. It might be a simple as treating `EntityDecoder[SomeExceptionWrapper\/T]` special and then an `EntityDecoder[T]` could be lifted up inside another `EntityDecoder`.

Jacoby6000: I'm not sure about special treatment of an inner type like that.... How about how http4s does it using `handle`?  

The final product would look something like

```scala
GET / "foo" / "bar" ^ EntityDecoder[Baz] |>> { 
  ... 
} handleDecodeErrors { 
  ... 
}
```

However, I think this won't work with the existing architecture, because of the way routes are composed using some side-effects in the background.

core/src/main/scala/org/http4s/rho/bits/PathTree.scala
```diff
@@ -88,13 +88,11 @@ private[rho] object PathTree {
...
         Leaf { (req, pathstack) =>
-          RuleExecutor.runRequestRules(req, c.router.rules, pathstack).map{ i =>
-            parser.decode(req, false).run.flatMap(_.fold(e =>
-              e.toHttpResponse(req.httpVersion),
-              { body =>
-                // `asInstanceOf` to turn the untyped HList to type T
-                route.action.act(req, (body :: i).asInstanceOf[T])
-              }))
+          RuleExecutor.runRequestRules(req, c.router.rules, pathstack).flatMap{ i =>
+            parser.parse(req).run.map{ body =>
```
Jacoby6000: Perhaps instead of my changes, we allow people to optionally pass in an error handler during the construction of a `CodecRouter`.  The error handler would then be used here in a `fold`. Basically just like the old way, but without a hard-coded failure case.

bryce-anderson: I like that idea: being able to pass in a handler while maintaining the
ability to continue the current behavior.

bryce-anderson:
@Jacoby6000: wow, that got a lot simpler, didn't it? In general, I like where this is going. 👍

Jacoby6000:
I'm going to close this because it's been awhile since I've touched it.  Most of what I have done is probably irrelevant now, and I don't have plans to get back to it any time soon.

PR #132, opened 2016-07-01
Expose the type tag of a CodecRouter

etorreborre:
@bryce-anderson this is necessary for me to implement the following function
```scala
  // create a PATCH route from the POST route, just changing the
  // Method and Response body
  def createPatchRoute[T <: HList](route: RhoRoute[T]): RhoRoute[T] =
    ???
```

In the implementation of this method I need to access the `CodecRouter` type tag and copy it to a modified `CodecRouter`. If I don't do this I get `Any` as the type of the body.

Note that the whole thing is a work-around because I was not able to abstract some code between similar POST and PATCH routes without losing the precise response types.

### DIFF
--- a/core/src/main/scala/org/http4s/rho/Router.scala
+++ b/core/src/main/scala/org/http4s/rho/Router.scala
@@ -11,11 +11,14 @@ import shapeless.{HNil, ::, HList}
 import shapeless.ops.hlist.Prepend
 
 sealed trait RoutingEntity[T <: HList] {
+  type Self <: RoutingEntity[T]
+
   def method: Method
   def path: PathRule
   def rules: RequestRule
 
-  def /:(path: TypedPath[HNil]): RoutingEntity[T]
+  def /:(path: TypedPath[HNil]): Self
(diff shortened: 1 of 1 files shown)


bryce-anderson:
I feel like there must be a better way to do this than exposing the `TypeTag`. Maybe changing the HTTP method is generally useful and we could provide a method on `RhoRoute` to do that?

etorreborre:
This is possibly useful indeed. I'll try to do that instead.

etorreborre:
This simplifies my code actually quite a bit. Thanks for the suggestion @bryce-anderson.

bryce-anderson:
@etorreborre: can we close this PR? Sounds like you've found a better option.

etorreborre:
@bryce-anderson the better option *is* in the PR :-). I amended it with a `withMethod` method to be able to set a new method on a `RoutingEntity` without having to expose the `TypeTag`.

bryce-anderson:
I don't know how I missed that, sorry! This LGTM, thanks!

PR #134, opened 2016-07-01
Added an object property to nest attributes on a definition

etorreborre:
The `ObjectProperty` is part of the Java Swagger model and is actually useful for me in order to avoid having to create and reference another definition when I can just nest attributes.


bryce-anderson:
👍 Thanks again!

PR #137, opened 2016-09-13
Add the possibility to specify a swagger 'file' type for a given response

etorreborre:
@bryce-anderson feel free to change the naming of the marker trait.

Also note that I changed slightly the primitive type check from using `=:=` to `<:<`. I think this is working ok but maybe you prefer to consider `file` as an entirely separate case and not treat it as "primitive" (it is a Swagger primitive type but not a Scala one).

### DIFF
--- a/swagger/src/main/scala/org/http4s/rho/swagger/package.scala
+++ b/swagger/src/main/scala/org/http4s/rho/swagger/package.scala
@@ -34,7 +34,7 @@ package object swagger {
         typeOf[java.lang.Double], typeOf[java.lang.Float], typeOf[BigDecimal],
         typeOf[java.lang.Byte], typeOf[java.lang.Boolean], typeOf[Number],
         typeOf[java.lang.Short], typeOf[Date], typeOf[Timestamp], typeOf[scala.Symbol],
-        typeOf[java.math.BigDecimal], typeOf[java.math.BigInteger], typeOf[SwaggerFileResponse])
+        typeOf[java.math.BigDecimal], typeOf[java.math.BigInteger])
     }
 
     private[swagger] val excludes = {
@@ -87,7 +87,7 @@ package object swagger {
       t <:< typeOf[Option[_]]
 
     def isPrimitive: Boolean =
-      Reflector.primitives.find(t <:< _).isDefined ||
+      Reflector.primitives.find(_ =:= t).isDefined ||
         Reflector.isPrimitive(t, Set(typeOf[Char], typeOf[Unit]))
 
     def isProcess: Boolean =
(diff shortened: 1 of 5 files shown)


swagger/src/main/scala/org/http4s/rho/swagger/package.scala
```diff
@@ -97,4 +97,6 @@ package object swagger {
       t <:< typeOf[Task[_]]
   }
 
+  /** marker trait for responses returning files */
+  trait SwaggerFileResponse
```
bryce-anderson: I make this comment having spent only a few minutes considering it, so please bear with me. 😄

Could we make this a case class with a type parameter and an `EntityEncoder` implementation that simply resolves the `EntityEcoder` for the wrapped type and forwards to it? That way we don't need to extend anything in the future and all existing `EntityEncoder`s should still work on the underlying type.

Something to the tune of:
```scala
final case class SwaggerFileResponse[T](value: T)

object SwaggerFileResponse {
  implicit def entityEncoder[T](implicit encoder: EntityEncoder[T]): EntityEncoder[SwaggerFileResponse[T]] = ???
}
```

bryce-anderson: PS: I think this would also let us keep `=:=`.

etorreborre:
@bryce-anderson see my fix-up, I think your suggestion is much better and I am not piggy-backing on the primitives code anymore.

swagger/src/main/scala/org/http4s/rho/swagger/package.scala
```diff
@@ -95,6 +95,8 @@ package object swagger {
 
     def isTask: Boolean =
       t <:< typeOf[Task[_]]
-  }
 
+    def isSwaggerFile: Boolean =
+      t <:< typeOf[SwaggerFileResponse[_]]
```
bryce-anderson: ffti: I suspect that this could be `=:=`.

bryce-anderson:
I like it! Only one suggestion, but I doubt it makes any difference either way so I intend to merge this PR sometime this evening.

swagger/src/main/scala/org/http4s/rho/swagger/package.scala
etorreborre: I tried that but the test doesn't pass and I don't really know why `=:=` doesn't work in that case.

bryce-anderson: Interesting! I wonder why that could be. I don't think it matters for the purposes of this PR.

bryce-anderson:
Thanks again @etorreborre!

Issue #140, opened 2016-09-22
Custom field serializers don't work with Option(al) fields


phile314-isska:
Example:
<pre>
  val swaggerFormats = DefaultSwaggerFormats.withFieldSerializers({
    case x if x =:= typeOf[DateTime] => AbstractProperty(`type`="string", format = Some("date-time"))
  })
</pre>
JSON Response Case Class:
<pre>
case class Project(
                    id:Option[Long],
                    name:String,
                    modified: Option[DateTime],
                    created: DateTime)
</pre>
Generated Swagger:
<pre>
{
"swagger":"2.0",
"info":{"version":"1.0.0","title":"My API"},
"paths":{
  "/projects":{
    "get":{
      "tags":["projects"],
      "summary":"List of all projects belonging to the authenticated user.",
      "operationId":"getProjects",
      "produces":["application/json"],
      "responses":{
        "200":{
          "description":"OK",
          "schema": {
            "type":"array","items":{"$ref":"#/definitions/Project"}
          }
        }
      }
    }
  }
},
"definitions":{
  "Project":{
    "required":["created","description","name"],
    "properties":{
      "name":{"type":"string"},
      "description":{"type":"string"},
      "id":{"type":"integer","format":"int64"},
      "modified":<b>{"type":"DateTime","description":"org.joda.time.DateTime"}</b>,
      "created":<b>{"type":"string","format":"date-time"}</b>
      },
    "description":"Project"
    }
  }
}
</pre>

bryce-anderson:
Have you had a chance to check PR #139?

bryce-anderson:
@phile314-isska, see #141. I'd appreciate you testing it with your use case before I merge it.

phile314-isska:
Just tested it, works perfectly :+1:

bryce-anderson:
Fixed by #141.

PR #141, opened 2016-09-24
Correctly use custom serializers with Option wrapped types

bryce-anderson:
Fixes issue #140

Problem

The Option type was being passed to DataType.fromType
which doesn't know about custom serializers.

Solution

Unwrap the Option wrapper and `TypeBuilder.typeToProperty` and
recurse.


phile314-isska:
Works fine for me.

PR #144, opened 2016-10-19
Use pretty-printing for swagger schema.

phile314-isska:
Makes schema much more readable and makes the line numbers from Swagger Editor error messages much more helpful.

As the Swagger-schema is normally only used for dev/documentation purposes, I think the increase in size shouldn't be a problem.


bryce-anderson:
I don't care about the readability bit (there are already browser plugins for that) but I do agree with the error message concern.

I'm going to delay merging this in a day it so just in case someone wants to object.

bryce-anderson:
Thanks @phile314-isska!

PR #146, opened 2016-10-24
Fix handling optional complex but AnyVal query parameters

jcranky:

PR #147, opened 2016-10-25
Don't encode case class AnyVals as RefProperty

jcranky:


jcranky:
This should be the last one for now, with this, rho-swagger works properly for us :)

rossabaker:
Release with this and #146, then?

bryce-anderson:
@rossabaker, sounds like a plan to me.

cc @etorreborre, might want to take a quick peak in cause you have objections.

etorreborre:
@bryce-anderson no objections I've already discussed this with @jcranky as we are on the same team :-).

jcranky:
@bryce-anderson thanks !

bryce-anderson:
v0.13.1 is off to central. Thanks again @jcranky!

jcranky:
thanks :D

PR #148, opened 2016-12-12
Update http4s version to 0.15.0

jcranky:


jcranky:
@rossabaker is this enough?

Issue #136, opened 2016-09-01
incompatibility of Scala 2.11.8


shengc:
Is it a known issue to the Lightbend ? If so, can you please provide the ticket number tracking the issue ? Our work has been stuck at 2.11.7 because of this :-(

bryce-anderson:
I don't know if it is or not. @rossabaker suggested bisecting scalac between 2.11.7 and 2.11.8, but that has been pretty low on my priority list.

rossabaker:
I don't think it's been minimized to the point that it is actionable by Lightbend.  @reactormonk was thinking about a bisector script for another problem.  He wanted a script that returned 0 or not to show the pattern.  Is that bisection still a relatively easy thing to write, as slow as it might be to run?

shengc:
I have been able to narrow down the issue to the following,

```scala
  object route extends RhoService {
    GET / "root" |>> { () ⇒ Ok("Hello") }
  }

  object route2 extends RhoService {
    val ok = Ok("Hello")
    GET / "root" |>> { () ⇒ ok }
  }
```

so with 2.11.8 `route` compiles, `route2` _does not_, but both of them compiles in 2.11.7. I had tried to turn on the implicit log to see how compiler resolves the implicits, but the output is sort of overwhelming to me. Can you pick up from here to see if this can be further minimized to something that Lightbend can act on ?

shengc:
what is weird is just by putting `implicitly[EntityEncoder[String]]`in context will solve all compile errors. If I remove the `implicit` keyword from `charSequenceEncoder`, compile errors go away except for the ones that does not take type parameter, such as `CONTINUE`, `REDIRECT`, etc. I think the implicit instances of `EntityEncoder` do have ambiguities from the perspective of 2.11.8, and some are not just found yet.

rossabaker:
I remember Bryce having a very difficult time organizing the priorities of the implicits in EntityEncoder to make it work.

This is unlikely, but I wonder if it does better with the SI2712-fix compiler plugin?  I hit some cases on the cats branch where implicits were ambiguous  and the plugin made it find the expected one.  I don't think that will work here, because I don't think this is an issue of kinds, but it might be worth a try.

shengc:
[...]

PR #149, opened 2016-12-13
Fixed issue #136 and added Scala-2.12 cross compile

zarthross:
Also updated Http4s dependency

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: scala
 scala:
-- 2.11.7
+- 2.11.8
+- 2.12.1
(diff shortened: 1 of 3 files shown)


zarthross:
Hmm, I might have not integrated the cross compile entirely correctly, but I think this is a start.

zarthross:
Well, I'm not travis expert, so not entirely sure why its failing to compile because it can't find Scala 2.10 when its supposed to be using 2.12.1, suggestions are welcome.

reactormonk:
The 2.10 was for SBT, fixed.

PR #152, opened 2016-12-23
Issue 150: Changed PathTree.splitPath to url decode path elements

zarthross:
Hey guys, let me know what you think of this solution.
I added some tests as well.


core/src/main/scala/org/http4s/rho/bits/PathTree.scala
```diff
@@ -58,14 +58,18 @@ private[rho] object PathTree {
...
     @tailrec
-    def go(i: Int, begin: Int): Unit = if (i < len) {
-      if (path.charAt(i) == '/') {
-        if (i > begin) buff += path.substring(begin, i)
-        go(i + 1, i + 1)
-      } else go(i + 1, begin)
-    } else {
-      buff += path.substring(begin, i)
+    def go(i: Int, begin: Int): Unit = {
+      def buffer(): Unit = buff += java.net.URLDecoder.decode(path.substring(begin, i), "UTF-8")
```
bryce-anderson: I think in the path we don't want to decode `+` as `space` which java.net.URLDecoder does. You could use the http4s `UrlCoding.decode` utility function. It's http4s private, but since rho lives in the http4s package it should work, although I'm not thrilled about it :unamused:. If we hadn't been so lucky, an alternative solution is to replace any '+' chars with "%2b" before doing the decoding (preferably before doing the slicing so its all done in one pass).

I'd prefer to inline `buffer` since its only used twice, but if you feel strongly about it, then its fine.

zarthross: Do you mean to manually inline it or to use the `@inline` annotation?

zarthross:
No rush on merging.  Just had some time around the holidays to do some hacking!

core/src/main/scala/org/http4s/rho/bits/PathTree.scala
bryce-anderson: Your last commit was pretty much exactly what I wanted, thanks!

core/test.sc
```diff
@@ -0,0 +1,21 @@
+import org.http4s.rho.{RhoService, _}
```
jcranky: What is this file?

zarthross: A file that shouldn't have gotten committed.  I'll remove it.

jcranky:
Now it looks good to me :)

bryce-anderson:
Thanks again @zarthross!

etorreborre:
Thanks @zarthross!

PR #153, opened 2016-12-23
Refactored examples to not use deprecated methods

zarthross:
Refactored examples to not use deprecated methods and also refactored them so the entry points look basically the same.


examples/src/main/scala/com/http4s/rho/swagger/demo/Main.scala
```diff
@@ -1,15 +1,28 @@
...
 import org.http4s.server.blaze.BlazeBuilder
+import org.http4s.server.{Server, ServerApp}
+import org.log4s.getLogger
+
+import scalaz.concurrent.Task
+
+object Main extends ServerApp {
+  private val logger = getLogger
+
+  val port = Option(System.getenv("HTTP_PORT"))
```
bryce-anderson: Nice touch :+1:

bryce-anderson:
Thanks @zarthross!

examples/src/main/scala/com/http4s/rho/swagger/demo/Main.scala
zarthross: Can't take the credit for that one, it was lifted directly from the existing Hal example.  Glad you like it though.

zarthross:
No problem!

PR #154, opened 2016-12-24
Added option to disable swagger routes in swagger

zarthross:
Personally, I thought it was a little weird that the swagger routes were showing up in swagger.  So I've added an option to disable this while leaving the default behavior the same.


bryce-anderson:
Thanks @zarthross!

phile314-isska:
I found that behavior strange as well, nice fix 👍 I think it would even make sense to not show the swagger routes by default?

zarthross:
@phile314-isska I personally think they should not be shown by default as well, but the code looked like it was put in that way intentionally, so I wanted to leave the the default behavior as it was.  I'll let @bryce-anderson change the default if they feel it should be changed.

bryce-anderson:
I'm not strongly opinionated about it and would be fine with a switch to
not shown by default.

PR #159, opened 2016-12-30
Hide swagger routes in swagger documentation by default.

phile314-isska:


zarthross:
Probably simmered long enough... Don't you think @bryce-anderson?

bryce-anderson:
Agreed.

PR #160, opened 2016-12-30
Clean up a few sections of the API

bryce-anderson:
We have a lot of API surface right now that should probably not be exposed. This commit starts to trim away the things that aren't necessary for public consumption.

Targets include:

* RouteExecutable: `runWith` is unnecessary and very inefficient.
* CompileService: pull ServiceBuilder into its own file and clean it up.
* Minor other changes.

This is a binary incompatible change, but shouldn't be controversial.

### DIFF
--- a/core/src/main/scala/org/http4s/rho/ServiceBuilder.scala
+++ b/core/src/main/scala/org/http4s/rho/ServiceBuilder.scala
@@ -1,0 +1,58 @@
+package org.http4s.rho
+
+import org.http4s._
+import org.log4s._
+import shapeless.HList
+
+import scala.collection.immutable.VectorBuilder
+
+/** CompileService which accumulates routes and can build a `HttpService` */
+final class ServiceBuilder private(internalRoutes: VectorBuilder[RhoRoute.Tpe]) extends CompileService[RhoRoute.Tpe] {
+  private val logger = getLogger
+
+  /** Turn the accumulated routes into an `HttpService`
+    *
+    * @param filter [[RhoMiddleware]] to apply to the collection of routes.
+    * @return An `HttpService` which can be mounted by http4s servers.
+    */
+  def toService(filter: RhoMiddleware = identity): HttpService =
+    CompileService.foldServices(internalRoutes.result(), filter)
+
+  /** Get a snapshot of the currently acquired routes */
+  def routes(): Seq[RhoRoute.Tpe] = internalRoutes.result()
+
+  /** Append the routes into this [[ServiceBuilder]]
+    *
+    * @param routes Routes to accumulate.
+    * @return `this` instance with its internal state mutated.
+    */
+  def append(routes: TraversableOnce[RhoRoute.Tpe]): this.type = {
+    internalRoutes ++= routes
+    this
+  }
+
+  /** Accumulate the [[RhoRoute]] into this [[ServiceBuilder]]
+    *
+    * This is the same as appending a the single route and returning the same route.
+    *
+    * @param route [[RhoRoute]] to compile.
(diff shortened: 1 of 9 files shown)


core/src/main/scala/org/http4s/rho/ServiceBuilder.scala
```diff
@@ -0,0 +1,58 @@
...
+    *
+    * This is the same as appending a the single route and returning the same route.
+    *
+    * @param route [[RhoRoute]] to compile.
+    * @tparam T `HList` representation of the result of the route
+    * @return The [[RhoRoute]] passed to the method.
+    */
+  override def compile[T <: HList](route: RhoRoute[T]): RhoRoute[T] = {
+    internalRoutes += route
+    route
```
zarthross: Any particular reason you removed the synchronized blocks around internalRoutes?

core/src/test/scala/org/http4s/rho/ApiTest.scala
```diff
@@ -1,22 +1,27 @@
...
+
 import scalaz.concurrent.Task
 import scalaz.stream.Process
-import scodec.bits.ByteVector
+import scalaz.{-\/, \/-}
 
 // TODO: these tests are a bit of a mess
 class ApiTest extends Specification {
+
+  def runWith[T <: HList, F](exec: RouteExecutable[T])(f: F)(implicit hltf: HListToFunc[T, F]): Request => Task[Response] = {
```
zarthross: The name 'runWith' doesn't read as well in Polish notation as it did in infix.  I suggest either rename it or maybe make it an extension method instead so the syntax stays the same.

core/src/test/scala/org/http4s/rho/RhoServiceSpec.scala
```diff
@@ -336,7 +337,7 @@ class RhoServiceSpec extends Specification with RequestRunner {
     }
   }
 
-  "RhoService concatonation" should {
+  "RhoService and method" should {
     "concatonate service" in {
```
zarthross: Was this supposed to be 'concatenate'?

core/src/main/scala/org/http4s/rho/ServiceBuilder.scala
bryce-anderson: All operations weren't synchronized so it was either expand it or remove it. Being a builder, it seemed unlikely to be used concurrently.

core/src/test/scala/org/http4s/rho/ApiTest.scala
bryce-anderson: I'm not concerned about the syntax of tests. Do you see a use case for this method on the `RouteExecutable` itself outside of testing in rho? I don't think its all that useful for testing outside of rho since its only tests that the route has assembled the right list of parameters for you, which should be enforced by the compiler anyway.
Perhaps I'm just being short sited. :smile:

core/src/test/scala/org/http4s/rho/RhoServiceSpec.scala
bryce-anderson: Yes. Good eye.

core/src/main/scala/org/http4s/rho/ServiceBuilder.scala
zarthross: Fair enough

core/src/test/scala/org/http4s/rho/ApiTest.scala
zarthross: Was thinking maybe useful for writing tests against rho routes for library users?

PR #161, opened 2017-01-10
Generate unique operation ids for paths having different parameters

etorreborre:
Without this PR we get errors on our Swagger spec with paths like
```
/bar
/bar/<id>
```

Note that the path parameters are now collected in the declaration order `L.121`. There are other places where we do the same kind of reversal during the Swagger construction.

Review @bryce-anderson 

cc @jcranky

### DIFF
--- a/swagger/src/main/scala/org/http4s/rho/swagger/SwaggerModelsBuilder.scala
+++ b/swagger/src/main/scala/org/http4s/rho/swagger/SwaggerModelsBuilder.scala
@@ -118,7 +118,7 @@ private[swagger] class SwaggerModelsBuilder(formats: SwaggerFormats) {
         case CaptureTail::xs           => PathParameter(`type` = "string", name = "tail...".some):: Nil
       }
 
-    linearizeStack(rr.path::Nil).map(go(_, Nil)).flatten
+    linearizeStack(rr.path::Nil).flatMap(go(_, Nil)).reverse
   }
 
   def collectBodyParams(rr: RhoRoute[_]): Option[BodyParameter] =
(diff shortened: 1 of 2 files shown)


bryce-anderson:
I believe there is something more ugly going on here. Specifically, we are passing in a `RhoRoute`, which is not necessarily strictly linear, and expecting to get a single path back out
For example, the following path:
```scala
val p1 = "foo1" / pathVar[String]("name1")  // we accept two sub parts
val p2 = "foo2" / pathVar[String]("name2")
"foo4" ** GET / (p1 || p2) |>> { _: String => "don't care which path you take" }
```
yields the operation-ids "getFoo2-name2-name1" and "getFoo1-name2-name1".

As an aside, I think that few if any folks get much use from `or` rules and while I think they were clever, I don't think they pull their weight considering the complexity they introduce.

I think this PR makes it more correct, but its still not right. What do you want to do? I might pick at this a bit more this evening, but if you don't see anything by morning, it may be a bit before I get to it.

etorreborre:
I was not aware of the `||` operator, maybe there's a similar `reverse` to apply somewhere we're dealing with `||`. In any case the thing I am really after is the uniqueness of operation ids.

bryce-anderson:
I don't think it's as simple as a missing reverse. A RhoRoute cab have
multiple paths, but a lot of the swagger generation code seems ambivalent
to that fact.
Do you want me to merge this PR? My hunch is that this is blocking you
right now.

etorreborre:
Yes please merge the PR :-)

etorreborre:
Hi @bryce-anderson could we please have a release with this in? Thanks.

bryce-anderson:
Should be on its way. Thanks for the contribs!

PR #162, opened 2017-01-12
See if travis finds this one...

bryce-anderson:


bryce-anderson:
Looks like Travis is on duty. I must be the one MIA.

PR #163, opened 2017-01-14
Fixed Swagger Output of Map types

zarthross:
The swagger output for Map's both straight up and within another case class was an array instead of a object. 
This PR should correct the issue.


bryce-anderson:
Wow! I'm really surprised that case was missed. This looks good to me. I'll merge it tomorrow unless someone objects.

bryce-anderson:
Thanks for another great PR @zarthross!

PR #164, opened 2017-01-15
Added QueryParam name and HeaderKey name to error messages.

zarthross:
When the query param extractors or header extracts fail, the
error messages now include the name and the value.

### DIFF
--- a/core/src/main/scala/org/http4s/rho/package.scala
+++ b/core/src/main/scala/org/http4s/rho/package.scala
@@ -55,7 +55,7 @@ package object rho extends Http4s with ResultSyntaxInstances {
                (implicit parser: QueryParser[T], m: TypeTag[T]): TypedQuery[T :: HNil] =
     paramR(name, {t =>
       if (validate(t)) None
-      else Some(BadRequest("Invalid query parameter: \"" + t + "\""))
+      else Some(BadRequest(s"""Invalid query parameter: "$name" = "$t""""))
     })
 
   /** Define a query parameter that will be validated with the predicate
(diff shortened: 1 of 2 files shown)


bryce-anderson:
I like this, but it has come up recently that sensitive request information may leak via logging headers etc. I believe we were already leaking the potentially sensitive bits, namely the value, but this might be a good time to discuss it. I don't think that this PR makes the situation any worse.

@rossabaker, do you have an opinion on this? I think you're more familiar with this concern than I am.

zarthross:
I agree, this PR wouldn't make the leaking of sensitive headers much worse, just puts a name to the value.  Perhaps what needs to be done is the addition of a `TransformRule` ( Spit-balling names, definitely need a better one here) to the `RequestAST` that takes the `ResultResponse[T]` from the `TypedHeader` tree, transforming the Failure.  We could then add to the `TypedHeader[T]` a `transform(f: ResultResponse[T] => ResultResponse[T])` method... or at a bare minimum to solve our current problem `onError(f: FailureResponse => Task[FailureResponse])` (Or `ResponseReason => Task[ResponseReason]`)

Then users that don't want their header leaked could add their own error message like this:
```
GET / "Foobar" >>> capture(MyHeader).onError(f=> FailureResponse.error("bad Request")) |>> ???
```

Alternatively we could add a `captureFold` (like the fold on `Either` ) method similar to `captureMap`.  This would likely be more idiomatic to rho, but it is starting to add alot of methods to the rho package object that is starting to look monadic on header, query and path....

bryce-anderson:
So, after not coming to any groundbreaking conclusions regarding logging sensitive information that, I'm inclined to merge it since it doesn't make things worse and we can address the security related concerns separately.

If there are no objections, I'll merge this tomorrow.

rossabaker:
Sorry, I missed the ping.

My concern with logging header values would be that they're isolated in a category that is easy to turn off, and that it's clear how to do so.  Many authentication mechanisms rely solely on TLS for security, but thoughtlessly ship the unencrypted headers off to the company log aggregator.

bryce-anderson:
@rossabaker, no worries. I agree with the concern, but I don't think this PR does anything to make the situation worse than it is now. Do you see any reason not to merge this PR as is?

I've opened up a ticket regarding the potential issues about blindly logging values. (#167)

rossabaker:
👍

bryce-anderson:
Thanks again, @zarthross!

PR #168, opened 2017-01-21
Generate Swagger docs from multiple `RhoService`s. Fixes #166

### LINKED ISSUE #166: It would be nice if swagger generation was more flexible
The case that comes to mind, brought up in the http4s gitter room by @ktonga, was that it's not possible to use http4s middleware on a subset of routes and still generate complete swagger docs because the current state of swagger generation requires all the routes to be available in one pass. I don't expect this to be difficult to add.

ktonga:
Factor out Swagger routes (/swagger.json) creation so it can be reused
when the Swagger Specification is an aggregate from multiple `RhoService`s

### DIFF
--- a/swagger/src/main/scala/org/http4s/rho/swagger/SwaggerSupport.scala
+++ b/swagger/src/main/scala/org/http4s/rho/swagger/SwaggerSupport.scala
@@ -27,13 +27,8 @@ object SwaggerSupport {
         routes ++ (if(swaggerRoutesInSwagger) swaggerRoutes else Seq.empty )
       )
 
-    lazy val swaggerRoutes: Seq[RhoRoute[_ <: HList]] = new RhoService {
-      lazy val response = Ok(Json.mapper().writerWithDefaultPrettyPrinter().
-        writeValueAsString(swaggerSpec.toJModel)).
-        putHeaders(`Content-Type`(MediaType.`application/json`))
-
-      "Swagger documentation" ** GET / apiPath |>> (() => response)
-    }.getRoutes
+    lazy val swaggerRoutes: Seq[RhoRoute[_ <: HList]] =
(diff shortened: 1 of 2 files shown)


swagger/src/main/scala/org/http4s/rho/swagger/SwaggerSupport.scala
```diff
@@ -48,5 +43,22 @@ object SwaggerSupport {
     val sb = new SwaggerModelsBuilder(swaggerFormats)
     routes.foldLeft(Swagger())((s, r) => sb.mkSwagger(apiInfo, r)(s))
   }
+
+  /**
+   * Create a RhoService with the route to the Swagger json
+   * for the given Swagger Specification
+   */
+  def createSwaggerRoutes(
```
bryce-anderson: I believe we only create one route, so maybe `createSwaggerRoute` is more appropriate. Returning a `RhoService` with a single route also feels strange, but I don't think there is a good alternative.

ktonga: Agreed, I tried to keep the old names unchanged. But I've already changed it.

ktonga: I think returning `RhoService` is the best option to maniupulate them as the same just as I did in the test case.

ktonga:
Yeah, I had the same thought, I was about to only expose a compose of `createSwagger` + `createSwaggerRoute`, that's the only thing really needed to aggregate routes and add the doc endpoint. But for some reason I decided that having both separated is kinda more flexible, what was the whole point of the change.

bryce-anderson:
Thanks @ktonga!

PR #171, opened 2017-03-08
Swagger descriptions for sum types

reactormonk:
According to [the `discriminator` spec](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#models-with-polymorphism-support). There's still two description fields - one in the `ComposedModel` and one in the `RefModel`. Not sure how to fix that.

Also, how should the discriminator field be passed in? Via an annotation?

### DIFF
--- a/swagger/src/main/scala/org/http4s/rho/swagger/TypeBuilder.scala
+++ b/swagger/src/main/scala/org/http4s/rho/swagger/TypeBuilder.scala
@@ -81,39 +81,42 @@ object TypeBuilder {
           models ++ generics ++ children
 
         case tpe@TypeRef(_, sym: Symbol, tpeArgs: List[Type]) if isSumType(sym) =>
+          // TODO parameterize typeVar
           // TODO promote methods on sealed trait from children to model
-          modelToSwagger(tpe, sfs).map({ m =>
-            // TODO parameterize
-            val typeVar = "type"
-            m
-              .copy(discriminator = Some(typeVar))
-              .copy(`type` = Some("object"))
-              .copy(properties =
-                m.properties + (typeVar -> AbstractProperty(`type` = "string", required = true)))
-          }).toSet.flatMap({ model =>
-            val ref = RefModel(model.id + "Ref", model.id2 + "Ref", model.id2)
+          modelToSwagger(tpe, sfs).map(addDiscriminator("type")).toSet.flatMap { model =>
+            val parent = RefModel(model.id + "Ref", model.id2 + "Ref", model.id2)
             val children =
-              sym.asClass.knownDirectSubclasses
-                .flatMap({sym =>
-                  val tpe = sym.asType.toType
-                  val submodels = go(tpe, alreadyKnown, known)
-                  modelToSwagger(tpe, sfs).map({ subtype =>
-                    ComposedModel(
-                      id = subtype.id,
-                      id2 = subtype.id2,
-                      description = subtype.description,
-                      allOf = List(ref, subtype)
-                    )
-                  })
-                })
+              sym.asClass.knownDirectSubclasses.flatMap( sub =>
+                  modelToSwagger(sub.asType.toType, sfs).map(composedModel(parent))
+                )
             alreadyKnown ++ Set(model) ++ children
-          })
+          }
 
         case e =>
           Set.empty
       }
 
     go(t, alreadyKnown, known)
+  }
+
+  private def addDiscriminator(typeVar: String)(model: ModelImpl): ModelImpl = {
+    model
+      .copy(
+        discriminator = Some(typeVar),
+        `type` = Some("object"),
+        properties =
+        model.properties + (typeVar -> AbstractProperty(`type` = "string", required = true))
+      )
+  }
+
+  private def composedModel(parent: Model)(subtype: Model): Model = {
+    ComposedModel(
+      id = subtype.id,
+      id2 = subtype.id2,
+      description = subtype.description,
+      allOf = List(parent, subtype),
+      parent = parent.some
+    )
   }
 
   private[this] val defaultExcluded =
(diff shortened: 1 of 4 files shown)


swagger/src/main/scala/org/http4s/rho/swagger/models.scala
```diff
@@ -3,6 +3,7 @@ package org.http4s.rho.swagger
 import io.swagger.{models => jm}
 
 import scala.collection.JavaConversions._
+import scala.collection.JavaConverters._
```
jcranky: Having both `Converters` and `Conversions` together feels very weird... perhaps we could use only `Converters`?

reactormonk:
Uh-oh, seems to break the existing handling of enums via `sealed trait` + `case object`. Not sure how to deal with that.

swagger/src/main/scala/org/http4s/rho/swagger/models.scala
rossabaker: I thought `JavaConversions` was deprecated.  I haven't used it in years.

swagger/src/main/scala/org/http4s/rho/swagger/TypeBuilder.scala
```diff
@@ -80,6 +80,35 @@ object TypeBuilder {
 
           models ++ generics ++ children
 
+        case tpe@TypeRef(_, sym: Symbol, tpeArgs: List[Type]) if isSumType(sym) =>
+          // TODO promote methods on sealed trait from children to model
+          modelToSwagger(tpe, sfs).map({ m =>
```
bryce-anderson: I'm generally not one to complain about styel, but can we remove the extra `(` in `.map({ m =>`? Same below on the `.flatMap` below. If you break this down into named methods the `flatMap` should disappear.

swagger/src/main/scala/org/http4s/rho/swagger/TypeBuilder.scala
```diff
@@ -93,10 +122,15 @@ object TypeBuilder {
   private[this] def isCaseClass(sym: Symbol): Boolean =
     sym.isClass && sym.asClass.isCaseClass && sym.asClass.primaryConstructor.isMethod
 
+  private[this] def isSumType(sym: Symbol): Boolean =
+    sym.asClass.isSealed && ! sym.asClass.knownDirectSubclasses.forall({ symbol =>
```
bryce-anderson: Remove space between `!` and `sym.` Remove bonus `(`.

swagger/src/main/scala/org/http4s/rho/swagger/TypeBuilder.scala
```diff
@@ -80,6 +80,35 @@ object TypeBuilder {
 
           models ++ generics ++ children
 
+        case tpe@TypeRef(_, sym: Symbol, tpeArgs: List[Type]) if isSumType(sym) =>
```
bryce-anderson: This whole section feels like it could be broken down nicely into two named methods (pretty much where you have the `flatMap` call), and would be much more readable that way.

swagger/src/main/scala/org/http4s/rho/swagger/models.scala
```diff
@@ -243,13 +243,13 @@ object models {
       o.setParameters(fromList(parameters.map(_.toJModel)))
       o.setResponses(fromMap(responses.mapValues(_.toJModel)))
       o.setSecurity(fromList(security.map { m =>
-        m.mapValues(xs => xs : java.util.List[String]) : java.util.Map[String, java.util.List[String]]
```
bryce-anderson: That is much nicer, thank you.

swagger/src/main/scala/org/http4s/rho/swagger/TypeBuilder.scala
reactormonk: Sure. I've developed the habit of putting both everywhere, call it defensive syntax.

swagger/src/main/scala/org/http4s/rho/swagger/TypeBuilder.scala
reactormonk: I've decided on slightly different cuts with less parameters passed.

swagger/src/main/scala/org/http4s/rho/swagger/TypeBuilder.scala
```diff
@@ -80,23 +82,64 @@ object TypeBuilder {
...
       }
 
     go(t, alreadyKnown, known)
   }
 
+  private def addDiscriminator(sym: Symbol)(model: ModelImpl): ModelImpl = {
+    val typeVar = sym.annotations
+      .withFilter( a => a.tpe <:< typeOf[DiscriminatorField] )
+      .flatMap(a => a.tree.children.tail.collect { case Literal(Constant(field: String)) => field } )
+      .headOption.getOrElse("type")
```
bryce-anderson: I think that there isn't a reasonable default for the discriminator field or else we wouldn't need the annotation. In that case it may be better to reject the model and log the reason for failure at info or maybe even warning level.

swagger/src/test/scala/org/http4s/rho/swagger/TypeBuilderSpec.scala
```diff
@@ -291,6 +297,17 @@ class TypeBuilderSpec extends Specification {
...
     }
+
+    "Build a model for sealed traits" in {
+      val ms = modelOf[Sealed]
+      ms.foreach(_.toJModel) // Testing that there are no exceptions
+      ms.size must_== 3
+      val Some(seal: models.ModelImpl) = ms.find(_.id2 == "Sealed")
+      seal.discriminator must_== Some("foobar")
+      val Some(foo: models.ComposedModel) = ms.find(_.id2 == "FooSealed")
+      val Some(fooRef) = foo.allOf.collectFirst({case ref: models.RefModel => ref})
```
bryce-anderson: Don't worry about this `({` if you don't end up making another commit, but I'd like to get rid of it if you do.

swagger/src/main/scala/org/http4s/rho/swagger/TypeBuilder.scala
bryce-anderson: Much cleaner, thank you.

swagger/src/main/scala/org/http4s/rho/swagger/TypeBuilder.scala
reactormonk: `type` is the default for most libraries in terms of sum encoding. [argonaut-shapeless](https://github.com/alexarchambault/argonaut-shapeless#custom-encoding-for-sealed-traits), [spray-json-shapeless](https://github.com/fommil/spray-json-shapeless/blob/master/src/main/scala/fommil/sjs/FamilyFormats.scala#L394). Circe doesn't have a default.

PR #178, opened 2017-04-20
Added type: "object" to generated object types

reactormonk:
Apparently that's the proper way to go.


bryce-anderson:
Do you mind pointing me to the section of spec that defines this? I'm sure its right, and I'll look it up later, but if you find the time before I do, I'd appreciate it. 😄

reactormonk:
There isn't but the validator here is a bit picky. It should be correct in all cases where there's a `properties` field, so it shouldn't hurt.

bryce-anderson:
Sounds good to me. Sorry for the delay: I need to work out a better notification mechanism, these keep slipping through.

reactormonk:
... I tried messing around with `sonatypeRelease`, but apparently it's not that easy. You still gotta release manually.

bryce-anderson:
Its released and tagged as 0.16.2.

PR #180, opened 2017-05-29
Update to Http4s 0.17.0-M2, Cats, and FS2

zarthross:
First crack at updating to Cats/Fs2.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: scala
 scala:
 - 2.11.8
-- 2.12.1
+- 2.12.2
 jdk: oraclejdk8
 sudo: false
 cache:
@@ -10,7 +10,4 @@ cache:
 env:
   global:
     secure: P9xSHdfi5NvRIzYKO9svLwN4l6BS2NENdDPrVapMAZH30S+JPi/tpuU8o96vTMnXILW1bnMtBkWG718yVWjGn4HrKuTD91lKUMZDLgK6+nKpmrtEMcimQbwInPgCWp/GMY7XhM10KSBoe/UF7y0vAT+a3rI24pyLKTZ582Ci3n4=
-  matrix:
-  - SCALAZ_VERSION=7.1.11
-  - SCALAZ_VERSION=7.2.8
 script: bash bin/travis
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: scala
 scala:
-- 2.11.8
+- 2.11.11
 - 2.12.2
 jdk: oraclejdk8
 sudo: false
--- a/core/src/test/scala/org/http4s/rho/bits/ResponseGeneratorSpec.scala
+++ b/core/src/test/scala/org/http4s/rho/bits/ResponseGeneratorSpec.scala
@@ -1,53 +1,52 @@
 package org.http4s.rho.bits
 
-import org.http4s.headers.{`Content-Type`, Location, `Transfer-Encoding`, `Content-Length`}
+import org.http4s.headers.{Location, `Content-Length`, `Content-Type`, `Transfer-Encoding`}
 import org.http4s._
 import org.specs2.mutable.Specification
-
-
+import scodec.bits.ByteVector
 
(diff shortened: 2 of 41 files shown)


rossabaker:
Neat.  What's going to be the versioning strategy for rho?  Is it going to skip over http4s-0.16, or should he minor version be bumped out two versions?

(For those not following http4s-core development closely, 0.16 is the last scalaz-stream based release.)

zarthross:
I should think, Rho releases be at least synced up to the minor version of Http4s, even if the patch releases are out of sync.  

I would be happy to do another PR for a Http4s 0.16.* release of Rho but Rho is already released as 0.16 tied to 0.15.* of Http4s.

.travis.yml
```diff
@@ -1,7 +1,7 @@
 language: scala
 scala:
 - 2.11.8
```
ChristopherDavenport: Should have 2.11.11

core/src/test/scala/org/http4s/rho/bits/ResponseGeneratorSpec.scala
```diff
@@ -1,53 +1,52 @@
...
-
+import scodec.bits.ByteVector
 
 class ResponseGeneratorSpec extends Specification {
   import ResponseGeneratorInstances._
 
   "ResponseGenerator" should {
     "Build a response with a body" in {
-      val result = Ok("Foo").run
+      val result = Ok("Foo").unsafeRun
```
ChristopherDavenport: Could this be a property test with an Arbitrary String? If not, some code smells can get fixed up after this is merged.

zarthross: IMHO, thats out of the scope of this PR, the purpose was to get Rho updated for Http4s.  I agree we need to clean up A LOT of the tests, but I didn't want to change any of the behaviors of the code in this PR.

ChristopherDavenport: Totally fair. Ill make a memo to myself to clean these up later

bryce-anderson:
I'm in the middle of looking, and am being pulled away, but I'll try to finish it tomorrow. Looks great so far @zarthross! I'm pleasantly surprised how little fallout there is, but maybe that is all in the last half. :)

etorreborre:
That looks good to me too. When there's a RC version of both http4s / rho with fs2/cats I will do a pass on upgrading our libraries and will report back.

zarthross:
Well... looks like fs2 is going to be migrating to cats-effect instead of Task (https://github.com/functional-streams-for-scala/fs2/blob/b08d08b7c680c8f376ef71ddfa8921685a249c72/docs/migration-guide-0.10.md)

So, looks like this work will have to be redone for FS2 v0.10.

bryce-anderson:
Looks like we're good to go here. How do we want to manage the release of this? I suspect merge this PR and consider versioning later?

etorreborre:
@bryce-anderson given the comment about the next fs2 version I think this should wait until fs2 gets moved to cats-effect.

bryce-anderson:
sgtm.

etorreborre:
Ok, I see, this is is following up on a normal http4s release anyway, there's on reason not to do it. So let's go.

eklavya:
Please make a release with this?

bryce-anderson:
Sorry, I'm away from civilization for another day or so but I'll get to it
soon unless someone beats me too it.

ZizhengTai:
A realease would be really helpful.

bryce-anderson:
@zarthross @etorreborre are we still planning on holding this until an official http4s/fs2 release?

etorreborre:
@bryce-anderson no, I think we should release.

zarthross:
I also think we should release it, but probably as a snapshot or milestone in line with the http4s version.

bryce-anderson:
Okay, I'll see what needs to be done this evening and try and get a release
out.

bryce-anderson:
This has been released as v0.17.0-M3. Should be in central soon.

PR #183, opened 2017-06-13
Change SwaggerFormats to final case class

### LINKED ISSUE #182: Should `SwaggerFormats` be sealed?
It looks like the only thing that extends `SwaggerFormats` is `DefaultSwaggerFormats`. Could you just keep the `DefaultSwaggerFormats` object around (and probably rename it to just `SwaggerFormats`) and remove the trait? Or perhaps open up the trait to extension?

As it stands, if I want to customize some of the Swagger output of my models, I'm not sure quite sure whether I'm supposed to be using `DefaultSwaggerFormats` and then calling `withSerializers` or whether there was some oversight and I'm supposed to be extending `SwaggerFormats` directly.

changlinli:
SwaggerFormats used to be a sealed trait which was only extended by a
single object, DefaultSwaggerFormats, leaving it ambiguous how custom
SwaggerFormats were meant to be created. This changes it to a case class
to make it more apparent how things can be extended.

Fixes #182

### DIFF
--- a/swagger/src/main/scala/org/http4s/rho/swagger/SwaggerFormats.scala
+++ b/swagger/src/main/scala/org/http4s/rho/swagger/SwaggerFormats.scala
@@ -36,20 +36,6 @@ object SwaggerFormats {
   val emptySerializers: PartialFunction[Type, Set[Model]] = PartialFunction.empty
 
   val emptyFieldSerializers: PartialFunction[Type, Property] = PartialFunction.empty
-
-  val defaultSwaggerFormats: SwaggerFormats =
-    SwaggerFormats(ignoreNothingOrNull orElse ignoreExistentialType, emptyFieldSerializers)
-
-  val emptySwaggerFormats: SwaggerFormats =
-    SwaggerFormats(emptySerializers, emptyFieldSerializers)
-
-  def withSerializers(serializer: PartialFunction[Type, Set[Model]]): SwaggerFormats = {
-    SwaggerFormats(serializer, emptyFieldSerializers)
-  }
-
-  def withFieldSerializers(fieldSerializer: PartialFunction[Type, Property]): SwaggerFormats = {
-    SwaggerFormats(emptySerializers, fieldSerializer)
-  }
 }
 
 
--- a/swagger/src/main/scala/org/http4s/rho/swagger/SwaggerFormats.scala
+++ b/swagger/src/main/scala/org/http4s/rho/swagger/SwaggerFormats.scala
@@ -4,45 +4,53 @@ import scala.reflect.runtime.universe._
 
 import models._
 
-sealed trait SwaggerFormats { self =>
+final case class SwaggerFormats(customSerializers: PartialFunction[Type, Set[Model]],
+                                customFieldSerializers: PartialFunction[Type, Property]) {
 
-  def customSerializers: PartialFunction[Type, Set[Model]] = PartialFunction.empty
-  def customFieldSerializers: PartialFunction[Type, Property] = PartialFunction.empty
+  def withSerializers(serializer: PartialFunction[Type, Set[Model]]): SwaggerFormats =
+    this.copy(customSerializers = serializer orElse this.customSerializers)
 
-  /** Construct a new SwaggerFormats with custom model serializers */
-  def withSerializers(s: PartialFunction[Type, Set[Model]]): SwaggerFormats = new SwaggerFormats {
-    override val customSerializers: PartialFunction[Type, Set[Model]] = s orElse self.customSerializers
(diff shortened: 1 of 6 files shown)


changlinli:
I'm not sure about keeping `withSerializers` and `withFieldSerializers` plural in the case that we're adding just a single type, but not a huge deal. Also this isn't as backwards-compatible as I might've hoped. Lemme know whatever other feedback you've got.

swagger/src/main/scala/org/http4s/rho/swagger/SwaggerFormats.scala
```diff
@@ -4,45 +4,53 @@ import scala.reflect.runtime.universe._
 
 import models._
 
-sealed trait SwaggerFormats { self =>
+final case class SwaggerFormats(customSerializers: PartialFunction[Type, Set[Model]],
```
bryce-anderson: This is exactly what I had in mind. Looks great.

swagger/src/main/scala/org/http4s/rho/swagger/SwaggerFormats.scala
```diff
@@ -4,45 +4,53 @@ import scala.reflect.runtime.universe._
...
-    case tpe if tpe =:= t => ms
+  def withFieldSerializers(t: Type, property: Property): SwaggerFormats = withFieldSerializers {
+      case tpe if tpe =:= t => property
   }
 }
 
-object DefaultSwaggerFormats extends SwaggerFormats {
-
-  type F = PartialFunction[Type, Set[Model]]
+object SwaggerFormats {
```
bryce-anderson: It looks like the big reason it feels less backwards compatibility is because of the loss of the object `DefaultSwaggerFormats`. I think we could get most of it back if we make an _instance_, `val DefaultSwaggerFormats` in the swagger package object. It looks like you've also codefied an `EmptySwaggerFormats`, which could simply be another instance.

swagger/src/main/scala/org/http4s/rho/swagger/SwaggerFormats.scala
```diff
@@ -4,45 +4,53 @@ import scala.reflect.runtime.universe._
...
+
+  val emptyFieldSerializers: PartialFunction[Type, Property] = PartialFunction.empty
+
+  val defaultSwaggerFormats: SwaggerFormats =
+    SwaggerFormats(ignoreNothingOrNull orElse ignoreExistentialType, emptyFieldSerializers)
+
+  val emptySwaggerFormats: SwaggerFormats =
+    SwaggerFormats(emptySerializers, emptyFieldSerializers)
+
+  def withSerializers(serializer: PartialFunction[Type, Set[Model]]): SwaggerFormats = {
```
bryce-anderson: I think that if we have default and empty instances these functions that shadow the class methods are no longer necessary and someone can simply do `EmptySwaggerFormats.withSerializers(...)`.

swagger/src/main/scala/org/http4s/rho/swagger/SwaggerSupport.scala
```diff
@@ -17,10 +17,10 @@ object SwaggerSupport {
    * representing the full API
    */
   def apply(
-    swaggerFormats: SwaggerFormats = DefaultSwaggerFormats,
-    apiPath: TypedPath[HNil] = "swagger.json",
-    apiInfo: Info = Info(title = "My API", version = "1.0.0"),
-    swaggerRoutesInSwagger: Boolean = false): RhoMiddleware = { routes =>
+             swaggerFormats: SwaggerFormats = SwaggerFormats.defaultSwaggerFormats,
```
bryce-anderson: Can we undo the indent change? I'm open to being convinced this is a good idea but haven't seen this style.

swagger/src/main/scala/org/http4s/rho/swagger/SwaggerFormats.scala
changlinli: Makes sense

swagger/src/main/scala/org/http4s/rho/swagger/SwaggerFormats.scala
changlinli: Yep also makes sense

swagger/src/main/scala/org/http4s/rho/swagger/SwaggerSupport.scala
changlinli: I have no idea why this happened. Yeah lemme fix it.

bryce-anderson:
@changlinli merged! Thanks for the PR!

Issue #14, opened 2014-07-22
Authorization


bryce-anderson:
It would be nice to support authorizations directly in rho. I'm not sure of HAL would benefit from it, but swagger would.
**Prior Art:**
- [Scalatra](http://scalatra.org/2.3/guides/http/authentication.html) (suggested to be not so good by @casualjim)
- By proxy, [Warden](https://github.com/hassox/warden/wiki)

zarthross:
@bryce-anderson Authorization has come up again in #158, and I too need some support for it.  I would love to help out but could use some guidance as to how it should be implemented.

Some random thoughts/note here: 
- It would appear that HAL doesn't do anything with Authentication/Authorization so no need to worry about it.  
- The swagger 'security' models appear to be implemented in  [rho/swagger/src/main/scala/org/http4s/rho/swagger/models.scala ](https://github.com/http4s/rho/blob/master/swagger/src/main/scala/org/http4s/rho/swagger/models.scala) but swagger middleware doesn't extract any auth info.
- Swagger supports 3 kinds of authentication: OAuth, Api Key, Basic Authentication
- Swagger supports 1 kind of authorization: Scopes

So where to start? 
1. Forget swagger for now and add support for HTTP4s Auth middleware as a stop gap until we can get proper authorization?
1. Create an AuthedRhoService that uses an AuthedRequest instead of a Request object?  The AuthedRhoService could specify the security definitions for the service and each request could have a auth and/or 'scopes' extractor on the Route to specify the required scopes or validate the api-key for that route?
1. Just add an extractor like 'Headers' and PathVars work now, but adds the authentication metadata  for swagger and a parameter to the Action that contains the User/API_key what have you.

Thoughs?

zarthross:
I'm somewhat partial to the 3rd one, since you could mix Authed and Unauthed routes in one service.  Unfortunately it wouldn't share much with how HTTP4s does authentication since we have to keep the metadata.

bryce-anderson:
In my minds eye, it would be super useful to allow whole services to be protected at once, but I also don't think this should be mandatory and am not sure how that information would be surfaced to the action.

For the route by route case, I imagine it reasonably easy to be able to make a construct to the tune of
```scala
val authenticate: RoutingEntity[T] => RoutingEntity[Auth::T] = ???

val authedRoute = authenticate ( GET / "thesecrets") |>> { auth: Auth => ??? }
```

[...]

PR #184, opened 2017-07-08
[WIP] AuthedContext implementation

eklavya:
An uninformed and naive attempt at implementing an AuthedRhoService for #14 
The code is very likely poor quality and needs many changes, I hope it's not too bad 😛 

The test is not working at the moment. Will need help with that.
Tested separately with a dummy API, seems to work with http4s authentication middleware.


eklavya:
One more point, the auth part is not visible to swagger (what headers or auth info is required).

bryce-anderson:
Have you seen the function
```scala
def genericRequestHeaderCapture[R](f: Request => ResultResponse[R]): TypedHeader[R :: HNil]
```
and the `>>>` operator? I think you could build a lot of what you need using those tools. That should make this much less extensive and fit in a bit better with the current dsl. In my minds eye, you end up with something like this
```scala
object UserAuth {
  case class User(name: String, id: UUID)

  implicit val userAuthInfo = new AuthInfo[User] {
     ...
  }
}

object Auth {
  def authRule[T: AuthInfo]: TypedHeader[T :: HNil] = 
    genericRequestHeaderCapture { request =>
      auth stuff...
    }
}

val myService = new RhoService {
  GET / "auth" >>> Auth.authRule[UserAuth.User] |>> { user: UserAuth.User =>
    whatever...
  }
}
```

eklavya:
Ok, yeah that would be a lot less everything 👅 
However, how can I mix that with AuthedService. The main issue is being able to interoperate with the authentication middleware of http4s.

bryce-anderson:
Sounds like I didn't fully understand what you're trying to do. I'll take another look at it this evening.

core/src/main/scala/org/http4s/rho/CompileService.scala
```diff
@@ -43,4 +43,20 @@ object CompileService {
...
+
+  /** Convert the `Seq` of [[RhoRoute]]'s into a `AuthedService`
+    *
+    * @tparam U AuthInfo type for this service
+    * @param routes `Seq` of routes to bundle into a service.
+    * @param filter [[RhoMiddleware]] to apply to the routes.
+    * @param ev [[AuthInfo]] instance for U
+    * @return An `AuthedService`
+    */
+  def foldAuthedServices[U](routes: Seq[RhoRoute.Tpe], filter: RhoMiddleware = identity)(implicit ev: AuthInfo[U]): AuthedService[U] = {
```
bryce-anderson: Why can't this live in `AuthedServiceBuilder`? Furthermore, it seems like it's only purpose is to add the entry to the attribute map so couldn't it be
```scala
val rhoService = foldServices(routes, filter)
Service.lift { case AuthedRequest(auth, req) =>
  rhoService(req.withAttribute ...)
}
```

core/src/main/scala/org/http4s/rho/AuthInfo.scala
```diff
@@ -0,0 +1,18 @@
+package org.http4s.rho
+
+/** Typeclass for authentication embedding and retrieval from [[org.http4s.Request]] attributes map
+  *
+  * @tparam U Authentication info type for ex: User(id: UUID...)
+  */
+trait AuthInfo[U] {
```
bryce-anderson: Why do we need to serialize and then deserialize the type `U` when stuffing `U` in and pulling it out of the attribute map? Why not just put type `U` in there?

core/src/main/scala/org/http4s/rho/AuthedRhoService.scala
```diff
@@ -0,0 +1,68 @@
...
+  /** Convert the [[RhoRoute]]'s accumulated into a `AuthedService` */
+  final def toService(filter: RhoMiddleware = identity): AuthedService[U] = serviceBuilder.toAuthedService(filter)
+
+  final override def toString: String = s"AuthedRhoService(${serviceBuilder.routes().toString()})"
+
+  final override def /:(prefix: TypedPath[HNil]): AuthedRhoService[U] = {
+    new AuthedRhoService(serviceBuilder.routes().map { prefix /: _ })(ev)
+  }
+
+  def withAuth[R](req: Request)(f: U => R): R = {
```
bryce-anderson: Why do we need a closure here? It doesn't do anything but extract `U`, so why not make it a simple getter:
```scala
def getUath[U](req: Request): U = ...
```

core/src/main/scala/org/http4s/rho/AuthInfo.scala
eklavya: Didn't realise attributeMap allows to add any T !

core/src/main/scala/org/http4s/rho/AuthedRhoService.scala
eklavya: No reason, just an artefact of my usages in the project 👅

core/src/main/scala/org/http4s/rho/CompileService.scala
eklavya: I just wanted to conform to the overall existing structure.

eklavya:
@bryce-anderson thanks for taking the time to review. I have addressed the review comments, any stupidity is purely unintentional 😄

core/src/main/scala/org/http4s/rho/AuthedServiceBuilder.scala
```diff
@@ -0,0 +1,39 @@
...
+import org.http4s._
+import org.http4s.rho.CompileService.foldServices
+import org.http4s.rho.bits.PathTree
+import org.log4s._
+import shapeless.HList
+
+import scala.collection.immutable.VectorBuilder
+
+/** CompileService which accumulates routes and can build a `AuthedService` */
+final class AuthedServiceBuilder[U] private(internalRoutes: VectorBuilder[RhoRoute.Tpe]) extends ServiceBuilder(internalRoutes) {
```
bryce-anderson: I don't think we need this class at all: the important part
```scala
def toAuthedService(filter: RhoMiddleware = identity): AuthedService[U] = {
  val rhoService = toService(filter)
  Service.lift { case AuthedRequest(authInfo, req) =>
    rhoService(req.withAttribute[U](authKey, authInfo))
  }
}
```
can be inlined into `AuthedRhoService`.

core/src/main/scala/org/http4s/rho/AuthedRhoService.scala
```diff
@@ -0,0 +1,65 @@
...
+  *     }
+  *   }
+  *
+  * }}}
+  *
+  * @tparam U authInfo type for this service.
+  * @param routes Routes to prepend before elements in the constructor.
+  *
+  */
+class AuthedRhoService[U](routes: Seq[RhoRoute[_ <: HList]] = Vector.empty)
```
bryce-anderson: I'm not convinced this needs to be a full DSL: all you really want is the `toAuthedService` and `getAuth[U]` functions which you could just use from an instance of this without having all the noise of another DSL class. This also helps to mitigate the `AttributeKey` issue described below.

core/src/main/scala/org/http4s/rho/AuthedRhoService.scala
```diff
@@ -0,0 +1,65 @@
...
+
+  final implicit protected def compileService: CompileService[RhoRoute.Tpe] = serviceBuilder
+
+  /** Create a new [[AuthedRhoService]] by appending the routes of the passed [[AuthedRhoService]]
+    *
+    * @param other [[AuthedRhoService]] whos routes are to be appended.
+    * @return A new [[AuthedRhoService]] that contains the routes of the other service appended
+    *         the the routes contained in this service.
+    */
+  final def and(other: AuthedRhoService[U]): AuthedRhoService[U] = new AuthedRhoService(this.getRoutes ++ other.getRoutes)
```
bryce-anderson: This is going to have problems with the referential equality of `AttributeKey`: when you define a route, you use the `getAuth` of the DSL instance where it is defined. However, the conversion of the routes to an `AuthedService` uses the key from the DSL instance performing the compilation, resulting in different keys.

core/src/main/scala/org/http4s/rho/ServiceBuilder.scala
```diff
@@ -7,7 +7,7 @@ import shapeless.HList
 import scala.collection.immutable.VectorBuilder
 
 /** CompileService which accumulates routes and can build a `HttpService` */
-final class ServiceBuilder private(internalRoutes: VectorBuilder[RhoRoute.Tpe]) extends CompileService[RhoRoute.Tpe] {
+class ServiceBuilder (internalRoutes: VectorBuilder[RhoRoute.Tpe]) extends CompileService[RhoRoute.Tpe] {
```
bryce-anderson: See the comments above about why we don't need to extend this. If you're going to change this file at all lets make the class private as well.

core/src/main/scala/org/http4s/rho/AuthedRhoService.scala
eklavya: I am not sure if I understand what you mean.

bryce-anderson: What I mean is that function composition gives you the vast majority of stuff you need. For example, the class below would work for you:
```scala
class AuthContext[T] {
  private val key = AttributeKey[U]("key")
  def getUser(req: Request): T = req.attributes.get(key).get
  def transform(rhoService: Service[Request, Response]): Service[AuthedRequest[U], Response] =
    Service.lift { case AuthedRequest(authInfo, req) =>
      rhoService(req.withAttribute[U](key, authInfo))
    }
}
```

Then you just need to have this instance's `getUser` method when you construct your `RhoService` so you can actually get your user, and then use the same instance to transform it into an `AuthedService`. Note that this isn't a rho specific tool. The only downside is that nobody is stopping you from forgetting the second part and then `getUser` will explode since it clearly won't find the user in the attribute map. That last part is the only thing that `AuthedRhoService` really buys us.

eklavya: Is the downside enough incentive to have a minimal DSL on AuthedRhoService?

eklavya: In fact I think, it would be very unpleasant (maybe even impossible) to do it without making it a DSL. You can't call getAuth inside a RhoService which you need to give to transform.

bryce-anderson: It's totally possible and not that bad:
```scala
object MyAuth extends AuthContext[User]

object MyService extends RhoService {
  import MyAuth._
  GET/ "hello" |>> (req: Request) => {
    val user = getAuth(req)
    ???
  }
}
```

eklavya: You are right !! Making changes.

eklavya:
The only reason for using RhoService instead of HttpService was just to remove one more toService call on user side. Calling get here removes some boilerplate on the user side. In my opinion it's not too bad since a wrong implementation (where service is not an AuthedService) would fail pretty soon in a project and is a done deal (you do it only once).

I don't completely agree that this is not rho specific because it's all about making rho work with the auth middleware of http4s. That said, this solves my purpose and it's ok if it's not merged. I did see two issues about this so maybe this is needed elsewhere too. But it's entirely your call :)

bryce-anderson:
I'm happy to merge it into rho and if others find utility outside, we can talk about moving it to http4s proper. I can merge it as is, or, if you're interested, we could integrate this with the rho dsl using a custom path rule:
```scala
final val auth: TypedHeader[U :: HNil] = rho.genericRequestHeaderCapture { req =>
    req.attributes.get(authKey) match {
      case Some(user) => SuccessResponse(user)
      case None => FailureResponse.error("Invalid auth configuration")
    }
  }
```

Then we can use it in the dsl like such:
```scala
object MyService extends RhoService {
  import MyAuth._

  GET +? param("foo", "bar") >>> auth |>> { (req: Request, foo: String, user: User) =>
    if (user.name == "Test User") {
      Ok(s"just root with parameter 'foo=$foo'")
    } else {
      BadRequest("This should not have happened.")
    }
  }
}
```

With even a little more work, that custom rule could probably include the metadata necessary for integration with the generated swagger docs, but that might be a significant amount of more work 😄.

If you're not interested in the dsl integration I can take care of it later. Thanks for the good work, and sorry this has taken so long but I think we've arrived at a really good commit.

eklavya:
Thanks @bryce-anderson 😄  
I can put more effort on this, sure.

eklavya:
I copy pasted your DSL code. I have no idea how to capture what http4s auth middleware is doing though.

bryce-anderson:
Merged! Thanks for the PR, and once again sorry for the lengthy back and forth but I think it paid off.

ZizhengTai:
Thanks guys. Will this be usable in a snapshot version perhaps?

eklavya:
@bryce-anderson thanks for your time and patience 😄 

A release with this would be really helpful.

Please let me know if you would like me to work on the swagger thing.

bryce-anderson:
Let me check when the next milestone release is for http4s proper: if it's close I'd like to wait for it.

PR #185, opened 2017-07-19
For #14 accept HttpService instead of RhoService

eklavya:
This is needed because we need to let user specify any RhoMiddleware they want.


eklavya:
@bryce-anderson I would also like to make swagger work without auth. Please let me know if there is a way for that.

bryce-anderson:
@eklavya I don't know of an easy way to make the routes all auth but the swagger.json.

PR #186, opened 2017-08-22
upgrade to http4s 0.17-RC1

eklavya:
Are we ok with unsafeFromLong or should that be handled?

### DIFF
--- a/core/src/main/scala/org/http4s/rho/bits/ResponseGenerator.scala
+++ b/core/src/main/scala/org/http4s/rho/bits/ResponseGenerator.scala
@@ -56,7 +56,7 @@ abstract class EntityResponseGenerator(val status: Status) extends ResponseGener
   def apply[A](body: A, headers: Headers)(implicit w: EntityEncoder[A]): Task[T[A]] = {
     w.toEntity(body).flatMap { case Entity(proc, len) =>
       val hs = len match {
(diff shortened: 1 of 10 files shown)


core/src/main/scala/org/http4s/rho/bits/ResponseGenerator.scala
```diff
@@ -56,7 +56,7 @@ abstract class EntityResponseGenerator(val status: Status) extends ResponseGener
   def apply[A](body: A, headers: Headers)(implicit w: EntityEncoder[A]): Task[T[A]] = {
     w.toEntity(body).flatMap { case Entity(proc, len) =>
       val hs = len match {
-        case Some(l) => (w.headers ++ headers).put(`Content-Length`(l))
+        case Some(l) => (w.headers ++ headers).put(`Content-Length`.unsafeFromLong(l))
```
rossabaker: I'm okay with these, because an Entity should never return a negative length, and if it does, we'll get a failed task.

(It does bring up that we can unsafely construct `Entities`, but that's a separate issue.  And once again, I wish we had a native unsigned long.)

bryce-anderson: I concur, doesn't look like a problem to me.

PR #187, opened 2017-08-24
Allow adding descriptions to parameters

eklavya:
I have changed the default version of Scala to 2.12.3 and set to 0.13.16 for fast compilation while developing. I hope that's alright.

I couldn't resolve method ambiguity for paramWithDesc. I tried adding an additional description: Option[String] = None to try and disambiguate but then I started receiving errors about missing implicit QueryParser. At that point I just changed the name to include "WithDesc".

### DIFF
--- a/core/src/main/scala/org/http4s/rho/AuthedContext.scala
+++ b/core/src/main/scala/org/http4s/rho/AuthedContext.scala
@@ -40,7 +40,7 @@ import org.http4s.rho.bits.{FailureResponse, SuccessResponse, TypedHeader}
 class AuthedContext[U] {
 
   /* Attribute key to lookup authInfo in request attributeMap . */
-  final private val authKey = AttributeKey[U]("authInfo")
+  final private val authKey = AttributeKey[U]
 
   /** Turn the [[HttpService]] into an `AuthedService`
     *
--- a/core/src/main/scala/org/http4s/rho/PathBuilder.scala
+++ b/core/src/main/scala/org/http4s/rho/PathBuilder.scala
@@ -59,7 +59,7 @@ final class PathBuilder[T <: HList](val method: Method, val path: PathRule)
     * @return a new [[PathBuilder]] that will capture a uri segment.
     */
   def /(symbol: Symbol): PathBuilder[String :: T] = {
-    val capture = PathCapture(symbol.name, StringParser.strParser, implicitly[TypeTag[String]])
+    val capture = PathCapture(symbol.name, None, StringParser.strParser, implicitly[TypeTag[String]])
     new PathBuilder(method, PathAnd(path, capture))
   }
 
--- a/core/src/main/scala/org/http4s/rho/bits/Metadata.scala
+++ b/core/src/main/scala/org/http4s/rho/bits/Metadata.scala
@@ -19,7 +19,7 @@ trait TextMetaData extends Metadata {
 }
 
 /** Metadata about a query rule */
-case class QueryMetaData[T](name: String, p: QueryParser[T], default: Option[T], m: TypeTag[T]) extends Metadata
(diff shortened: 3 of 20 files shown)


eklavya:
@bryce-anderson What do you mean? QueryMetadata has been changed to include description as well. Did you have something else in mind?

If we add another metadata type it would essentially be a copy of the current QueryMetadata with a field for description, no?

bryce-anderson:
@eklavya, first things first, you probably know more about the rho internals than I at this point so sorry for the situations where I don't make sense 😄. 

Have you noticed https://github.com/http4s/rho/blob/master/swagger/src/main/scala/org/http4s/rho/swagger/SwaggerModelsBuilder.scala#L181 ? I suspect not since it's only a very half baked bit of code that I don't think is even accessible through the dsl at this time, but it provides a glympse of what I had envisioned for adding metadata compositionally. All those type and API's could use some serious re-examination. As it is, I'm not sure whats going on with those subtypes. For example, why does a metadata class need a `Parser`? (maybe there is a reason but I'm not in a position to check)

**TL;DR:** I really appreciate you taking interest in this area of rho. Its clearly underdeveloped so this PR is a step up from the current state of things. I'd like to explore using the `MetaData` types in a more compositional way instead of adding fields to the base type, but if you strongly oppose I'm willing to yield and we can come back to it at a later time.

eklavya:
I had looked at it, but I couldn't figure it out completely.

Here's my (very likely wrong/uninformed) understanding of things.

I want to add descriptions to both ```pathVar```s and ```param```s.
I just started from both ends and passed the description parameter to connect them.

There seem to be a TextMetadata which is used in Route descriptions. Besides that there doesn't seem to be an active use of it. There are multiple fields in ```PathParameter``` and ```QueryParameter``` which need to be exposed in the DSL for the user to be able to use them. I am exposing description field in this PR. This is added for ```pathVar``` in ```PathCapture``` and for ```param``` in ```QueryMetaData```. In the future also if any other field needs to be exposed it should be added to these case classes in my opinion.

eklavya:
That said, I would very much like to improve things and make them compositional. What do you have in mind?

I would also like your help in getting rid of "withDesc" in the method names and somehow let overloading do it's job.

bryce-anderson:
I took another look at the `Metadata` tooling and it's not looking as simple as I'd hoped: the current restructuring of the AST in the `SwaggerModelsBuilder` isn't very fantastic right now. The reason you can't overload the `param` method is that the description signature ends up looking like the overload that takes a default:
```scala
def param[T](name: String, description: String, validate: T => Boolean)
def param[T](name: String, default: T, validate: T => Boolean)
```
I don't think we can do it directly during param construction without the different method name. The best way would be with an operator, but that means either using the implied knowledge that there is a `MetaRule` underneath or making the `MetaCons` AST work better.

So, what do you think? I'd prefer not to put the DSL through a bunch of chaos so we should probably pick a way and go with it.

eklavya:
> The reason you can't overload the param method is that the description signature ends up looking like the overload that takes a default.

Yeah, I was wondering if there was a trick to making it work somehow. I had tried adding an additional parameter (description: Option[String] = None) to the one taking default but that started throwing errors about missing implicits and also it was an ugly hack. Anyways, we can keep "withDesc" in the name. Or should we make it "paramD" or something for less keystrokes?

So is this PR good to go then?

bryce-anderson:
Yeah, I'd say we should just move forward with this. Is there a way we can test this?

eklavya:
I also renamed ```paramWithDesc``` to ```paramD``` for finger health. I hope it's ok.

bryce-anderson:
Thats fine. Thanks for the PR, merging it now.

PR #190, opened 2017-09-28
Allow FailureResponses to be created with entities

jcranky:
This is necessary for more complex responses, like if you want to use
problem json.


jcranky:
I also wanted to add a test but I'm not sure where to put it. `ResultResponseSpec` doesn't seem to be the right place for it, but perhaps I'm wrong.

bryce-anderson:
I'll yield to your judgement on the testing. Fwiw, I'd be willing to merge this without a test.

jcranky:
@bryce-anderson if you don't mind merging as is, lets do it then :)

PR #193, opened 2017-10-19
Upgrade to http4s 0.18

jcranky:
In summary, replace all `Task`s with `F`s, and add `F` everywhere where
the new http4s version requires.


jcranky:
As stated in the title, this is still a WIP. Any feedback is already appreciated though :)

jcranky:
This is not compiling yet, and I'm getting several of those:

```scala
could not find implicit value for parameter F: cats.Monad[Nothing]
```

I guess I screwed up some implicits, but not sure yet where.

bryce-anderson:
Wow, thats a lot of work! I don't have time to review this this weekend, but I can try and take a look next week if you think its ready.

As for the implicits, they are pretty tough without abstracting over `F[_]`.

jcranky:
@bryce-anderson :)

I got some ideas from @rossabaker and other people in gitter, lets see when I can have this PR at least compiling. There is quite a bit changes to do still though. I didn't touch the tests at all yet, for example.

jcranky:
I'm a bit stuck. For now, I just want to make this compile, but the `Monad[Nothing]` that I mentioned above is not going away. Any ideas?

rossabaker:
It looks like you found your way around the `Monad[Nothing]`.  Do you need help on the outstanding errors, or are you still working on it?

jcranky:
@rossabaker yes, I got some help in gitter for that :)

One place where I could use some help is the old `TaskMatcher` - it now looks like this:

```scala
  implicit def fMatcher[F[_], R](implicit F: FlatMap[F], r: ResultMatcher[F, R]): ResultMatcher[F, F[R]] = new ResultMatcher[F, F[R]] {
    override def encodings: Set[MediaType] = r.encodings
    override def resultInfo: Set[ResultInfo] = r.resultInfo

    override def conv(req: Request[F], t: F[R])(implicit F: Monad[F], w: EntityEncoder[F, F[R]]): F[Response[F]] = F.flatMap(t)(r.conv(req, _))
  }
```

and the error is:

```scala
[error] [12] core/src/main/scala/org/http4s/rho/bits/ResultMatcher.scala:446:
[error]      Cannot convert from R to an Entity, because no EntityEncoder[F, R] instance could be found.
[error]          override def conv(req: Request[F], t: F[R])(implicit F: Monad[F], w: EntityEncoder[F, F[R]]): F[Response[F]] = F.flatMap(t)(r.conv(req, _))
```

Now, I'm not sure that method makes sense anymore... or does it?

ChristopherDavenport:
`w: EntityEncoder[F, F[R]]` that signature gives me pause. Why is the `EntityEncoder[F, F[R]]` as opposed to `EntityEncoder[F, R]`

rossabaker:
I agree with @ChristopherDavenport on the compiler error.  `EntityEncoder[F, R]` compiles in my head.

I am not highly experienced with rho as either a user or a developer, but I think the whole method still makes sense, so routes can return `F[A]` and be flattened if we know what to do with an `A`.  This might also be a wildly ignorant statement.

jcranky:
Yes, I agree that `EntityEncoder[F, F[R]]` smells really weird. I'll experiment with `EntityEncoder[F, R]`.

jcranky:
Simply changing the type there seems to do the trick. I guess I was confused we that error mixed up with others before :)

jcranky:
I was on vacation but now I'm getting back to this :)
I rebased the PR on top of the current master. My current challenge is a few compilation errors happening in `PathTree`, like:

```scala
[error] [1] core/src/main/scala/org/http4s/rho/bits/PathTree.scala:52:
[error]     type mismatch;
[error]      found   : scala.collection.immutable.Map[Nothing,Nothing]
[error]      required: Map[String,org.http4s.rho.bits.PathTree.MatchNode[F]]
[error]     Note: Nothing <: String, but trait Map is invariant in type A.
[error]     You may wish to investigate a wildcard type such as `_ <: String`. (SLS 3.2.10)
[error]     Error occurred in an application involving default arguments.
[error]       def apply[F[_]](): PathTree[F] = new PathTree[F](MatchNode[F](""))
```

But I don't understand why it is finding a `Map[Nothing, Nothing]` there :/

rossabaker:
Maybe that earlier advice to flatten out `org.http4s.EntityEncoder[F,F[R]]` has come back to haunt us?

jcranky:
@rossabaker maybe, I'm still not totally sure about how to deal with that yet

jcranky:
@bryce-anderson @rossabaker @ChristopherDavenport @eklavya and whoever else is interested: this is not a 100% done but is already in a reviewable state - any feedback is very much appreciated.

Also, there are a few TODOs left, which is why this is not a 100% done, and any help with those would be welcome.

build.sbt
```diff
@@ -71,7 +71,13 @@ lazy val `rho-examples` = project
...
   .dependsOn(`rho-swagger`, `rho-hal`)
 
-lazy val compileFlags = Seq("-feature") //, "-Xlog-implicits")
+lazy val compileFlags = Seq(
+  "-feature",
+  "-deprecation",
+  "-Ypartial-unification",
+  "-language:higherKinds",
+  "-language:existentials",
+  "-language:implicitConversions") //, "-Xlog-implicits")
```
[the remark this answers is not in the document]
jcranky: No, it was like that before, so I left if there. I could remove though :)

build.sbt
```diff
@@ -98,7 +104,8 @@ lazy val buildSettings = publishing ++
       logbackClassic % "test"
     ),
     libraryDependencies ++= specs2,
-    libraryDependencies += `scala-reflect` % scalaVersion.value
+    libraryDependencies += `scala-reflect` % scalaVersion.value,
+    addCompilerPlugin("org.spire-math" % "kind-projector" % "0.9.4" cross CrossVersion.binary)
```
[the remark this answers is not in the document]
jcranky: I have no idea :) It turned out I was using it only in two places, so.... removed.

core/src/test/scala/ApiExamples.scala
```diff
@@ -108,83 +105,79 @@ class ApiExamples extends Specification {
...
       /// end_src_inlined
 
       /// src_inlined Decoders
-      new RhoService {
+      new RhoService[IO] {
         // Using decoders you can parse the body as well
-        POST / "postSomething" ^ UrlForm.entityDecoder |>> { m: UrlForm =>
-          Ok(s"You posted these things: $m")
+        POST / "postSomething" ^ UrlForm.entityDecoder[IO] |>> { m: UrlForm =>
+          Ok[IO](s"You posted these things: $m")
```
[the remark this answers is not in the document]
jcranky: I agree with that, that type in there is very awkward. I'll experiment a bit around that idea, thanks :)

jcranky:
@bryce-anderson @rossabaker in rho there is a comment that states:

`// UseProxy has been deprecated in RFC7231 so it is omitted.`

`UseProxy` seems to be present in `http4s` itself though, so this is a bit inconsistent.

core/src/test/scala/ApiExamples.scala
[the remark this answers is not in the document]
jcranky: Done :)

build.sbt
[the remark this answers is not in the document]
bryce-anderson: Yeah, I noticed that you didn't add it. Feel free to remove it, or not. :)

core/src/test/scala/ApiExamples.scala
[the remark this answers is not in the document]
bryce-anderson: Awesome! I'm glad it worked out.

build.sbt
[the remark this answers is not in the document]
jcranky: removed :)

core/src/main/scala/org/http4s/rho/bits/PathAST.scala
```diff
@@ -106,7 +106,7 @@ object PathAST {
 
   case class PathMatch(s: String) extends PathOperation
 
-  case class PathCapture(name: String, description: Option[String], parser: StringParser[_], m: TypeTag[_]) extends PathOperation
+  case class PathCapture[F[_]](name: String, description: Option[String], parser: StringParser[F, String], m: TypeTag[_]) extends PathOperation
```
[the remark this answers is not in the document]
jcranky: I'm not sure I get your point, but the parsers got a bit tricky because they need stuff that depends on `F`.

core/src/main/scala/org/http4s/rho/bits/PathTree.scala
```diff
@@ -75,65 +74,70 @@ private[rho] object PathTree {
...
-          RuleExecutor.runRequestRules(req, c.router.rules, pathstack).map{ i =>
-            parser.decode(req, false).value.flatMap(_.fold(e =>
-              e.toHttpResponse(req.httpVersion),
-              { body =>
-                // `asInstanceOf` to turn the untyped HList to type T
-                route.action.act(req, (body :: i).asInstanceOf[T])
-              }))
-          }
-        }
+        ??? // TODO: fix
```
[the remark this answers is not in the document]
jcranky: This and the `???` in the function below it are the two main places where I mentioned in some comment above that I could use some help - I'm not sure how to make that compile :/

jcranky:
@bryce-anderson regarding what is missing, it is just a few specific places like the `???` you just pointed.

jcranky:
@bryce-anderson I just fixed one more issue and double checked the project. Those two TODOs on PathTree are the only missing piece right now :) - I might have to fix other things after that though, because until those two are done I don't really know if the tests are passing :/

bryce-anderson:
Sounds good. I'll try to find some more time for reviewing this evening.
I'm curious what the result of the test will be, but I suspect there won't
be any show stoppers from that code.

jcranky:
Thank you @danielkarch for the help!

jcranky:
For the record, the missing piece now is an implementation of `isEffect` that makes sense and works.

core/src/main/scala/org/http4s/rho/bits/PathTree.scala
```diff
@@ -311,7 +323,7 @@ private[rho] object PathTree {
   private def mergeMatches(m1: Map[String, MatchNode], m2: Map[String, MatchNode]): Map[String, MatchNode] =
     mergeMaps(m1, m2)(_ merge _)
 
-  private def mergeLeaves(l1: Map[Method,Leaf], l2: Map[Method,Leaf]): Map[Method,Leaf] =
+  private def mergeLeaves(l1: Map[Method, Leaf], l2: Map[Method, Leaf]): Map[Method, Leaf ] =
```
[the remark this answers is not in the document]
jcranky: done

core/src/main/scala/org/http4s/rho/bits/QueryParser.scala
```diff
@@ -1,32 +1,36 @@
...
-    override def collect(name: String, params: Params, default: Option[Option[A]]): ResultResponse[Option[A]] = {
-      val defaultValue = default.getOrElse(None)
+  implicit def optionParse[A](implicit F: Monad[F], p: StringParser[F, A]) = new QueryParser[F, Option[A]] {
+    override def collect(name: String, params: Params, default: Option[Option[A]]): ResultResponse[F, Option[A]] = {
       params.get(name) match {
         case Some(Seq(value, _*)) =>
           p.parse(value) match {
-            case SuccessResponse(value) => SuccessResponse(Some(value))
-            case other => other.asInstanceOf[ResultResponse[Option[A]]]
+            case SuccessResponse(successValue) => SuccessResponse[F, Option[A]](Option[A](successValue))
```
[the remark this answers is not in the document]
jcranky: I think I was having problems with type inference before the code was stable, this can now be simplified. done.

core/src/main/scala/org/http4s/rho/Result.scala
```diff
@@ -68,70 +68,41 @@ sealed case class Result[
...
-    override def withStatus(status: Status): Self =
+    override def withStatus(status: Status)(implicit F: Functor[F]): Self =
       Result(r.resp.copy(status = status))
 
-    override def attemptAs[T](implicit decoder: EntityDecoder[T]): DecodeResult[T] = {
-      val t: Task[Either[DecodeFailure, T]] = r.resp.attemptAs(decoder).value
-      EitherT[Task, DecodeFailure, T](t)
+    override def attemptAs[U](implicit F: FlatMap[F], decoder: EntityDecoder[F, U]): DecodeResult[F, U] = {
+      val t: F[Either[DecodeFailure, U]] = r.resp.attemptAs(F, decoder).value
+      EitherT[F, DecodeFailure, U](t)
```
jmcardon: doesn't this return `EitherT[F, DecodeFailure, U]` before you call `.value`? 

If so why unwrap then re-wrap?

core/src/main/scala/org/http4s/rho/RouteExecutable.scala
```diff
@@ -1,27 +1,24 @@
...
   */
-trait RouteExecutable[T <: HList] extends TypedBuilder[T] { exec =>
+trait RouteExecutable[F[_], T <: HList] extends TypedBuilder[F, T] { exec =>
 
   /** `Method` of the incoming HTTP `Request` */
   def method: Method
 
-//  /** Create a [[RhoRoute]] from this [[RouteExecutable]] with the provided converters */
-  def makeRoute(action: Action[T]): RhoRoute[T]
+  /** Create a [[RhoRoute]] from a given this [[Action]] */
```
jmcardon: from a given [[Action]]

core/src/main/scala/org/http4s/rho/Router.scala
```diff
@@ -28,63 +30,63 @@ sealed trait RoutingEntity[T <: HList] {
...
 
-  override def >>>[T1 <: HList](v: TypedHeader[T1])(implicit prep1: Prepend[T1, T]): Router[prep1.Out] =
+  override def >>>[T1 <: HList](v: TypedHeader[F, T1])(implicit prep1: Prepend[T1, T]): Router[F, prep1.Out] =
     Router(method, path, AndRule(rules, v.rule))
 
-  override def makeRoute(action: Action[T]): RhoRoute[T] = RhoRoute(this, action)
+  override def makeRoute(action: Action[F, T]): RhoRoute[F, T] = RhoRoute(this, action)
 
-  override def decoding[R](decoder: EntityDecoder[R])(implicit t: TypeTag[R]): CodecRouter[T, R] =
+  override def decoding[R](decoder: EntityDecoder[F, R])(implicit F: Functor[F], t: TypeTag[R]): CodecRouter[F, T, R] =
```
jmcardon: bound here seem unused

swagger/src/main/scala/org/http4s/rho/swagger/package.scala
```diff
@@ -106,10 +71,10 @@ package object swagger {
         Reflector.isPrimitive(t, Set(typeOf[Char], typeOf[Unit]))
 
     def isStream: Boolean =
-      t <:< typeOf[Stream[Task, _]]
+      t <:< typeOf[Stream[_]]
```
jmcardon: you removed the fs2 stream import by accident I think.

`def isStream: Boolean = t <: typeOf[Stream[F, _]]` or `Stream[_, _]`

swagger/src/main/scala/org/http4s/rho/swagger/Reflector.scala
```diff
@@ -0,0 +1,28 @@
...
+import java.sql.Timestamp
+import java.util.Date
+
+import org.http4s.rho.bits.ResponseGenerator.EmptyRe
+
+object Reflector {
+  import scala.reflect.runtime.universe._
+
+  private[swagger] val primitives = {
+    Set[Type](typeOf[String], typeOf[Int], typeOf[Long], typeOf[Double],
```
jmcardon: So  just for the sake of performance, couldn't you make this an array? Set access and iteration is slow, and the fact it's a set is pretty much an implementation detail, especially considering that it's private so it doesn't have to matter. See [haoyi's post](http://www.lihaoyi.com/post/BenchmarkingScalaCollections.html#sets-and-maps-are-slow)

That aside, you can also make it more performant by removing the need of concatenation with the empty set as a default as well. That can be a different call for the cases where that isn't used (which seems like most of them)

I'm nitpicking here btw. Premature optimization, etc, so feel free to ignore.

core/src/main/scala/org/http4s/rho/bits/QueryParser.scala
```diff
@@ -1,34 +1,38 @@
 package org.http4s
 package rho.bits
 
-import scala.language.higherKinds
+import cats.Monad
+import org.http4s.rho.bits.QueryParser.Params
 
 import scala.annotation.tailrec
 import scala.collection.generic.CanBuildFrom
+import scala.language.higherKinds
```
jmcardon: this import can go I think.

swagger/src/main/scala/org/http4s/rho/swagger/package.scala
jcranky: True, I noticed that already. I just didn't fix it because I was only able to make it work by hard coding `IO` instead of F. I think this is basically the same problem as with `isEffect` - which also works if I use `IO`. (the code compiles with F, but tests fail)

core/src/main/scala/org/http4s/rho/Result.scala
```diff
@@ -1,12 +1,12 @@
...
 package rho
 
+import cats.{FlatMap, Functor, Monad}
 import cats.data.EitherT
 
-import fs2.Task
-
 /** A helper for capturing the result types and status codes from routes */
 sealed case class Result[
+F[_],
```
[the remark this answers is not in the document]
jcranky: Ha! This is tricky, but this `+` is github stuff, and not from the code itself :)

core/src/main/scala/org/http4s/rho/Result.scala
```diff
@@ -68,70 +68,41 @@ sealed case class Result[
...
 
-trait ResultSyntaxInstances {
-
-  implicit class ResultSyntax[T >: Result.TopResult <: BaseResult](r: T) extends ResponseOps {
+trait ResultSyntaxInstances[F[_]] {
+  implicit class ResultSyntax[T >: Result.TopResult[F] <: BaseResult[F]](r: T) extends ResponseOps[F] {
     override type Self = T
 
-    override def withStatus(status: Status): Self =
+    override def withStatus(status: Status)(implicit F: Functor[F]): Self =
```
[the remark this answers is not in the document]
jcranky: This comes from _http4s_ itself, and since it is an override, we have to keep it.

core/src/main/scala/org/http4s/rho/Result.scala
jcranky: Good catch, will fix.

core/src/main/scala/org/http4s/rho/RouteExecutable.scala
jcranky: thanks, will fix

core/src/main/scala/org/http4s/rho/Result.scala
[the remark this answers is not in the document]
jmcardon: LOL. My bad. I didn't notice it because it was right beside the new line :man_facepalming:

core/src/main/scala/org/http4s/rho/Result.scala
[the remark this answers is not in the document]
jmcardon: ah ok. no problem.

core/src/main/scala/org/http4s/rho/Router.scala
jcranky: another case that comes from `http4s`, a bit more indirectly - it is necessary in line 90 for the `orElse`

core/src/main/scala/org/http4s/rho/bits/QueryParser.scala
jcranky: true, will fix

swagger/src/main/scala/org/http4s/rho/swagger/Reflector.scala
jcranky: This was already like this in 'rho', it just appears as a change because I moved it, though I don't fully remember why. I noted your commented anyway and will gladly have a new PR addressing it after this one is done.

core/src/main/scala/org/http4s/rho/AuthedContext.scala
```diff
@@ -47,22 +49,23 @@ class AuthedContext[U] {
...
   }
 
   /* Get the authInfo object from request. */
-  def getAuth(req: Request): U = {
+  def getAuth(req: Request[F]): U = {
     req.attributes.get[U](authKey).get
   }
 
-  final val auth: TypedHeader[U :: HNil] = rho.genericRequestHeaderCapture { req =>
+  def auth(): TypedHeader[F, U :: HNil] = RhoDsl[F].genericRequestHeaderCapture[U] { req =>
```
etorreborre: Does this really need a `()`?

danielkarch:
I just wanted to summarize the issue with `isEffect` in `swagger/package.scala`. This used to be 
```scala
    def isTask: Boolean =
      t <:< typeOf[Task[_]]
```
where `t` is a `Type`. 
The implementation cannot work for `isEffect`, because effect types do not have a subtyping relationship with `cats.effect.Effect`. We would have to know that, given some `Type` at runtime, there is an implicit `Effect` instance for that type. I do not think this is possible. Does anyone here have an idea for a workaround?

core/src/main/scala/org/http4s/rho/AuthedContext.scala
jcranky: You can still access it without the parens.

core/src/main/scala/org/http4s/rho/AuthedContext.scala
```diff
@@ -37,7 +39,7 @@ import org.http4s.rho.bits.{FailureResponse, SuccessResponse, TypedHeader}
   *
   * @tparam U authInfo type for this service.
   */
-class AuthedContext[U] {
+class AuthedContext[F[_]: Monad, U] extends FailureResponseOps[F] {
```
ChristopherDavenport: Should this be `U, F[_]: Monad`?

core/src/main/scala/org/http4s/rho/bits/ResultMatcher.scala
```diff
@@ -345,103 +351,108 @@ object ResultMatcher {
       INSUFFICIENTSTORAGE,
       LOOPDETECTED,
       NOTEXTENDED,
-      NETWORKAUTHENTICATIONREQUIRED]): Task[Response] = Task.now(r.resp)
+      NETWORKAUTHENTICATIONREQUIRED])(implicit F: Monad[F]): F[Response[F]] = F.pure(r.resp)
```
ChristopherDavenport: Why Monad over Applicative if we only use `pure`?

core/src/main/scala/org/http4s/rho/AuthedContext.scala
jcranky: Why?

core/src/main/scala/org/http4s/rho/bits/ResultMatcher.scala
jcranky: Because this extends the `ResultMatcher` trait, and some of the other implementations use `flatMap`.

core/src/main/scala/org/http4s/rho/AuthedContext.scala
ChristopherDavenport: Partial Unification Reasons, but I could be wrong.

core/src/main/scala/org/http4s/rho/bits/ResultResponse.scala
```diff
@@ -1,84 +1,92 @@
 package org.http4s.rho.bits
 
+import cats.data.OptionT
+import cats.{Applicative, Functor, Monad}
```
bryce-anderson: I don't think we need `Applicative`.

core/src/main/scala/org/http4s/rho/bits/ResultResponse.scala
```diff
@@ -1,84 +1,92 @@
...
+    this match {
+      case SuccessResponse(t) => OptionT(ev(t).map(Option.apply))
+      case NoMatch()          => OptionT.none[F, Response[F]]
+      case FailureResponse(r) => OptionT(r.toResponse.map(Option.apply))
     }
 }
 
 /** Failure to match a route */
-case object NoMatch extends RouteResult[Nothing]
+case class NoMatch[F[_]]() extends RouteResult[F, Nothing]
```
bryce-anderson: I'm a little queasy of making this a case class since it will be a pretty common result when walking the `PathTree`. A hacky solution (which doesn't fix the part about it being a case class but does fix the root allocation concern) is to make a singleton and cast it to the right type:
```scala
/** Failure to match a route */
case class NoMatch[F[_]] private() extends RouteResult[F, Nothing]

object NoMatch {
  private[this] val singleton = new NoMatch[Nothing]
  // use `get` to not allocate a new `NoMatch`
  def get[F[_]]: NoMatch[F] = singleton.asInstanceOf[NoMatch[F]]
}
```

swagger/src/main/scala/org/http4s/rho/swagger/Reflector.scala
bryce-anderson: Generally speaking, I didn't care much about this code as its run once on startup, so as long as its not mega slow, it doesn't really matter. That said, an array is probably better.

core/src/main/scala/org/http4s/rho/bits/ResultResponse.scala
danielkarch: How about putting 

```scala
private val noMatch: RouteResult[F, Nothing] = NoMatch[F]()
```

into `PathTreeOps`? This would allocate it once whenever a `PathTree[F]()` is created, but there would be no need to cast.

core/src/main/scala/org/http4s/rho/AuthedContext.scala
jcranky: @ChristopherDavenport what would break? how can we test? @danielkarch his comment is related to a specific case in http4s where the params order had to be changed - afaik it happened in a single place there

core/src/main/scala/org/http4s/rho/bits/ResultResponse.scala
jcranky: @bryce-anderson I plan to send a new PR later cleaning up imports all over the place - I'm sure this is not the only place where this leftover happened :)

jcranky: @danielkarch true, will do

core/src/main/scala/org/http4s/rho/bits/ResultResponse.scala
jcranky: I'll experiment with @danielkarch 's idea - I would prefer to avoid casting as much as possible

core/src/main/scala/org/http4s/rho/bits/ResultResponse.scala
danielkarch: Any reason not to use  `-Ywarn-unused:imports` and `-Xfatal-warnings`?

core/src/main/scala/org/http4s/rho/bits/ResultResponse.scala
jcranky: ok, commited @danielkarch 's suggestion

core/src/main/scala/org/http4s/rho/bits/ResultResponse.scala
jcranky: @danielkarch that is exactly what I'll do in the new PR I mentioned I'll send ;)

bryce-anderson:
Again, great work @jcranky, thanks for tackling this. Did you want to get this released asap, or do you want to wait for a few more things?

jcranky:
I'm not in a hurry, although a milestone release could be useful to let people try this out and give feedback :)

jcranky:
Maybe we should have squashed that merge, but too late now XD

PR #194, opened 2017-10-20
Allow swagger routes with trailing slashes in Rho Swagger

andimiller:
This allows people to define a route with a final section of `/ ""` to insist that there's a slash on the end.

### DIFF
--- a/swagger/src/main/scala/org/http4s/rho/swagger/SwaggerModelsBuilder.scala
+++ b/swagger/src/main/scala/org/http4s/rho/swagger/SwaggerModelsBuilder.scala
@@ -97,6 +97,7 @@ private[swagger] class SwaggerModelsBuilder(formats: SwaggerFormats) {
     def go(stack: List[PathOperation], pathStr: String): String =
       stack match {
         case Nil                          => if(pathStr.isEmpty) "/" else pathStr
+        case PathMatch("")::Nil           => pathStr + "/"
         case PathMatch("")::xs            => go(xs, pathStr)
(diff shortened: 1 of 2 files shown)


andimiller:
This could also be implemented as a new class like `CaptureTail` called something like `TrailingSlash` with a symbol of `$` or similar.

andimiller:
As this PR is right now, the only people who'd see an extra trailing slash would be those who've intentionally added a `/ ""` onto the end of their routes, the alternative would be to require some kind of syntax eg. `/ $` which I don't feel is worth it.

jcranky:
As long as this is a non-breaking API change, seems ok to me - which seems to be the case. I tried to find some test that would guarantee this and could find one - perhaps we could add one test checking that a trailing slash is _not_ added when it shouldn't? Or is this covered by some test that I didn't see?

andimiller:
@jcranky Not adding extra trailing slashes to existing behaviour should be covered by the other tests that already existed in that file, eg. this one https://github.com/andimiller/rho/blob/b68646fc5988ef78a5361d96ffb901221e034a58/swagger/src/test/scala/org/http4s/rho/swagger/SwaggerSupportSpec.scala#L59-L71

jcranky:
@andimiller cool, how about adding a path ending with slash to that list ?

andimiller:
@jcranky sure I've just pushed extra tests to check mixing trailing/nontrailing endpoints

jcranky:
@rossabaker yes, looks good, thanks @andimiller

bryce-anderson:
👍 Thanks @andimiller!

Issue #196, opened 2017-10-28
Build with SBT 1.0


zarthross:
We should get the build scripts up to date with SBT 1 now that its released.

zarthross:
I've been working on doing these updates at work recently, so when I get a chance I'll try and get this repo building with it as well.  I've already taken steps to doing this with #195.

PR #198, opened 2017-11-13
Build with sbt 1.0 (#196)

jfraudeau:
I was able to build with sbt 1.0, 
Works :
 - publishing to GH Pages
 - Build the library
 - Unit tests

Not tested :
 - Publishing the project

### DIFF
--- a/project/build.properties
+++ b/project/build.properties
@@ -1,1 +1,1 @@
-sbt.version=1.0.2
+sbt.version=1.0.3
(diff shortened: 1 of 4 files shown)


project/build.properties
```diff
@@ -1 +1 @@
-sbt.version=0.13.16
+sbt.version=1.0.2
```
zarthross: Why did you not choose "1.0.3"?

jfraudeau: I simply didn't notice it was a out

jfraudeau:
Bumped the version

PR #199, opened 2017-11-17
security-scopes

TomLous:
Added way to define security scopes to RhoRoute:
using ^^ syntax to prefix a Map[String, List[String]] of security scopes

eg:

`Map("auth0_jwk"->List("admin")) ^^ "Description" ** GET / "foo" |>> { () => Ok("bar") }`

### DIFF
--- a/swagger/src/main/scala/org/http4s/rho/swagger/SwaggerModelsBuilder.scala
+++ b/swagger/src/main/scala/org/http4s/rho/swagger/SwaggerModelsBuilder.scala
@@ -157,7 +157,6 @@ private[swagger] class SwaggerModelsBuilder(formats: SwaggerFormats) {
     linearizeStack(rr.path::Nil).flatMap(go(_, None)).headOption
   }
 
-
   def collectSecurityScopes(rr: RhoRoute[_]): List[Map[String, List[String]]] = {
 
     def go(stack: List[PathOperation], summary: Option[Map[String, List[String]]]): Option[Map[String, List[String]]] =
@@ -178,7 +177,6 @@ private[swagger] class SwaggerModelsBuilder(formats: SwaggerFormats) {
 
     linearizeStack(rr.path::Nil).flatMap(go(_, None))
   }
-
 
   def collectOperationParams(rr: RhoRoute[_]): List[Parameter] =
     collectPathParams(rr) ::: collectQueryParams(rr) ::: collectHeaderParams(rr) ::: collectBodyParams(rr).toList
@@ -203,7 +201,6 @@ private[swagger] class SwaggerModelsBuilder(formats: SwaggerFormats) {
 
         case MetaRule(rs, m: TextMetaData)::xs =>
           go(rs::Nil).map(_.withDesc(m.msg.some)) ::: go(xs)
-
 
         case MetaRule(a, _)::xs => go(a::xs)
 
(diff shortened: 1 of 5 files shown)


swagger/src/main/scala/org/http4s/rho/swagger/SwaggerModelsBuilder.scala
```diff
@@ -157,6 +157,27 @@ private[swagger] class SwaggerModelsBuilder(formats: SwaggerFormats) {
...
+    def go(stack: List[PathOperation], summary: Option[Map[String, List[String]]]): Option[Map[String, List[String]]] =
+      stack match {
+        case PathMatch("")::Nil                => go(Nil, summary)
+        case PathMatch(s)::xs                  => go(xs, summary)
+        case PathCapture(id, _, parser, _)::xs => go(xs, summary)
+        case CaptureTail::xs                   => summary
+
+        case MetaCons(_, meta)::xs =>
+          meta match {
+            case RouteSecurityScope(secScope) => secScope.some
```
bryce-anderson: From the look of the function it seems that we potentially want to continue searching until the end, accumulating any data, otherwise we can eliminate the `summary` parameter.

swagger/src/main/scala/org/http4s/rho/swagger/SwaggerModelsBuilder.scala
```diff
@@ -157,6 +157,27 @@ private[swagger] class SwaggerModelsBuilder(formats: SwaggerFormats) {
...
+  def collectSecurityScopes(rr: RhoRoute[_]): List[Map[String, List[String]]] = {
+
+    def go(stack: List[PathOperation], summary: Option[Map[String, List[String]]]): Option[Map[String, List[String]]] =
+      stack match {
+        case PathMatch("")::Nil                => go(Nil, summary)
+        case PathMatch(s)::xs                  => go(xs, summary)
+        case PathCapture(id, _, parser, _)::xs => go(xs, summary)
+        case CaptureTail::xs                   => summary
+
+        case MetaCons(_, meta)::xs =>
```
bryce-anderson: Could this be dramatically by matching against
```
case MetaCons(_, RouteSecurityScope(secScope))::xs => .. whatever ..
case _::xs => go(xs, summary)
case Nil => summary
```

TomLous: Good point. I just copied from `collectSummary` without daring to touch/change too much.
Fixed now

swagger/src/main/scala/org/http4s/rho/swagger/SwaggerModelsBuilder.scala
TomLous: Also good point. I reduced unnecessary checks and removed summary.

rossabaker:
Seems reasonable to me.  There's a merge conflict now.

TomLous:
I rebased & merged so the conflict should be gone. Thanks for reviewing the code @rossabaker & @bryce-anderson

swagger/src/main/scala/org/http4s/rho/swagger/package.scala
```diff
@@ -25,6 +25,20 @@ package object swagger {
       new PathBuilder(builder.method, PathAST.MetaCons(builder.path, RouteDesc(description)))
   }
 
+  /** Scopes carrier for specific routes */
+  case class RouteSecurityScope(definitions: Map[String, List[String]]) extends SecurityScopesMetaData
+
+  /** Add support for adding documentation before a route using the ## operator */
```
ChristopherDavenport: I'm not really certain what this is referencing. It mentions `##` and then introduces `^^`.

TomLous: Correct. Typo. Thanks for catching it. I will fix immediately

PR #200, opened 2017-11-21
swagger-middleware-extension

TomLous:
The current implementation of the SwaggerSupport Middleware hid/didn't expose many of the available swagger configurations, like:
host, schemes, basePath, consumes, produces, security, securityDefinitions & vendorExtensions

Also The OAuth2Definition for security required a didn't allow for vendorExtensions and required tokenUrl.
So added OAuth2VendorExtensionsDefinition not to break compatibility.


TomLous:
Hi is there something I should do to notify a Reviewer or do I need to break it down more?

jcranky:
cc @bryce-anderson / @rossabaker

PR #201, opened 2017-11-21
Upgrade to 0.18 using IO

eklavya:
Making rho effect polymorphic is proving to be very very difficult. This PR uses IO everywhere and is on http4s 0.18

If somebody wants to work on making Rho effect polymorphic, maybe this can serve as a starting point.

What are your thoughts on having a Rho-io or something?


eklavya:
There is already #193

PR #209, opened 2018-01-09
Do not use void in swagger. Fixes #208

### LINKED ISSUE #208: rho should not use `void` in swagger
Hi.
I was looking for a way implement an endpoint that return no entity although the response code it returns allows presence of an entity (it works fine for 204 and other codes forbidden to have an entity). It turned out that:
```
PUT / "nothing" / "unit" |>> { () => Created(()) }
```
produces `swagger.json` with:
```
"/nothing/unit" : {
  "put" : {
    "tags" : [ "nothing" ],
    "operationId" : "putNothingUnit",
    "responses" : {
      "201" : {
        "description" : "Created",
        "schema" : {
          "type" : "void"
        }
      }
    }
  }
},
```
which contains `"type" : "void"`.
There's no `void` type in [swagger specification](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#data-types). IMO the `swagger.json` should contain no schema as the [specification suggests](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#response-object): "If this field does not exist, it means no content is returned as part of the response.":
```
"/nothing/unit" : {
  "put" : {
    "tags" : [ "nothing" ],
    "operationId" : "putNothingUnit",
    "responses" : {
      "201" : {
        "description" : "Created",
      }
    }
  }
},
```

RafalSumislawski:
With this PR empty responses are represented as Unit (the EmptyRe have been removed) and are not limited to response codes forbidding response entity. `OK(())` will work, and result in a correct swagger.json.

In other places where Unit/Void/Nothing/Null may be used (though most are ridiculous) rho will represent them as complex types. Unfortunately in some cases that will still result in a swagger.json not conforming to the specification. This is due to the limitation that parameters located in path, query or header must be of simple types. This is a separate issue, concerning not only Unit/Void/Nothing/Null but any complex type.

The "void" type have been completely removed.


core/src/main/scala/org/http4s/rho/bits/ResponseGenerator.scala
```diff
@@ -22,17 +19,6 @@ sealed trait ResponseGenerator {
   def status: Status
 }
 
-object ResponseGenerator {
```
bryce-anderson: This makes me happy!

jcranky:
@bryce-anderson I'll have a look later today :)

jcranky:
Sorry for the delay, looks good to me

bryce-anderson:
Thanks again for the PR @RafalSumislawski!

RafalSumislawski:
You're welcome. Thanks for merging it.

PR #211, opened 2018-01-20
Enable unused warnings and fatal warnings compiler flags

jcranky:


swagger/src/main/scala/org/http4s/rho/swagger/SwaggerSupport.scala
```diff
@@ -46,11 +46,10 @@ abstract class SwaggerSupport[F[_]](implicit F: Monad[F], etag: WeakTypeTag[F[_]
   }
 
   /**
-    * Create the swagger model for a set of routes
+    * Create the swagger model for a set of routes.
     */
   def createSwagger(
       swaggerFormats: SwaggerFormats = DefaultSwaggerFormats,
-      apiPath: TypedPath[F, HNil] = TypedPath(PathMatch("swagger.json")),
```
jcranky: @bryce-anderson The only change that is a bit "weird" is this one: the `apiPath` is gone, because it was not being used. Is this a problem? What was the original intention here?

bryce-anderson: I think that it originally was to allow changing the endpoint for the generated swagger JSON. Obviously that got lost along the way (if it ever worked), so this is fine with me 👍.

jcranky: There is the `createRhoMiddleware` function that seems to achieve that, so the purpose itself is still there :)

PR #217, opened 2018-02-09
parsers for Date and UUID types

eklavya:
for capturing path variables.

### DIFF
--- a/core/src/main/scala/org/http4s/rho/bits/StringParser.scala
+++ b/core/src/main/scala/org/http4s/rho/bits/StringParser.scala
@@ -2,6 +2,7 @@ package org.http4s
 package rho.bits
 
 import java.text.SimpleDateFormat
+import java.time.Instant
 import java.util.{Date, UUID}
 
 import cats.Monad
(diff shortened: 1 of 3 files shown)


core/src/test/scala/org/http4s/rho/bits/StringParserSpec.scala
```diff
@@ -51,5 +54,15 @@ class StringParserSpec extends Specification {
     "parse non-short" in {
       new ShortParser[IO]().parse("12345678") must haveClass[FailureResponse[IO]]
     }
+
+    "parse date" in {
```
jcranky: Should we test the failure cases as well ?

core/src/main/scala/org/http4s/rho/bits/StringParser.scala
```diff
@@ -82,6 +114,8 @@ object StringParser {
   implicit def intParser[F[_]]: IntParser[F] = new IntParser[F]()
   implicit def longParser[F[_]]: LongParser[F] = new LongParser[F]()
   implicit def shortParser[F[_]]: ShortParser[F] = new ShortParser[F]()
+  implicit def datePArser[F[_]]: DateParser[F] = new DateParser[F]()
```
Daenyth: `PA` here probably wants to be `Pa`

bryce-anderson:
Thanks @eklavya!

eklavya:
Thanks @bryce-anderson @jcranky @ChristopherDavenport :)

PR #224, opened 2018-02-16
Fix examples with different response codes

DStranger:
Widening response types to `F[BaseResult[F]]` loses information
about the actual responses.
We can make this compile by making `F` covariant.


jcranky:
I was bothered with the `.widen` indeed. I just ask myself if we want to use `F[+_]` as well. I would say it is less bad, but would be better to find a way to not need either :(

DStranger:
@jcranky I agree, I just don't see another way of preserving the type information. It was relying on `Task` being covariant before.

jcranky:
@DStranger would you mind to also do the infix style commented by @bryce-anderson  ?

PR #225, opened 2018-02-26
Handle Collection[A] in body parameter

eklavya:
Right now if the body is a List[A] or some collection, the generated swagger lists the parameter as List«A».
This PR fixes that and now we correctly encode it as an ArrayProperty.


jcranky:
Is it worth writing a test for that?

eklavya:
That would require a decoder for some List[A] which I think would require adding an external dependency (http4s-circe or something). So I didn't do it.

eklavya:
as it turns out a working decoder is not required 😉

PR #228, opened 2018-03-24
Fix issue 218 Conflicting import with Cats Implicits

zarthross:
Fixed the `diverging implicit expansion` error by using an Aux pattern on the `MaybeIsWritable` to hint the compiler  the correct F for implicit resolution


core/src/test/scala/ApiExamples.scala
```diff
@@ -28,7 +29,7 @@ class ApiExamples extends Specification {
         val pathPart1 = GET / "hello"
 
         pathPart1 / "world" |>> { () => Ok("Hello, world!") }
-        pathPart1 / "you"   |>> { () => Ok("Hello, you!") }
+        pathPart1 / "you"   |>> { () => Ok("Hello, you!".pure[IO]) }
```
jcranky: Why do we need this?

bryce-anderson:
@zarthross, do you mean that you added so that we have a reason to import `cats.implicits._`? If so, I think that's fine, though I'd add that as code comment since it does look out of place.

zarthross:
That is exactly why I added it.  I'll see if I can do something better/ more clear

zarthross:
Added an explicit test, and move the `pure` in API examples to somewhere already using an `IO`.  Does this address you concerns?

core/src/test/scala/org/http4s/rho/bits/HListToFuncSpec.scala
```diff
@@ -25,5 +25,14 @@ class HListToFuncSpec extends Specification {
       val req = Get("/route1")
       checkOk(req) should_== "foo"
     }
+
+    // Tests issue 218 https://github.com/http4s/rho/issues/218
+    "Work with cats.implicits" in {
+      import cats.implicits._
+      new RhoService[IO] {
+        GET / "route1" |>> { () => Ok("foo".pure[IO]) }
```
jcranky: Can we add a comment to say that the `pure[IO]` is here just for the import? My concern is people thinking that this is somewhat necessary.

zarthross: Done.

jcranky: thanks :D

PR #233, opened 2018-05-04
Remove unused import

jcranky:


jcranky:
ok, I got crazy in this one, ignore

PR #234, opened 2018-06-11
GET shouldn't have a body

nightscape:


bryce-anderson:
Change looks good, but can we retitle it? While it's poor form and uncommon to have a GET request with a body, it is not prohibited by the spec (though it is implicitly discouraged).
https://tools.ietf.org/html/rfc7231#section-4.3.1

nightscape:
Ok, I changed the title from `can't` to `shouldn't`.

PR #236, opened 2018-08-30
Upgrade to http4s 0.19 and friends

Igosuki:
Upgraded to work with http4s's latest 0.19.0-M2 (HttpRoutes) and fs2 1.0.0-M4 (removal of Segment, StreamApp) cats 1.2.0 and effects RC3 (Timer, ContextShift).


zarthross:
I think we can close this in favor of #253.

zarthross:
Thanks for your work @Igosuki!

PR #253, opened 2018-10-06
0.19

chuwy:
This is based on https://github.com/http4s/rho/pull/236 and leaves @Igosuki's work almost untouched, but:

* bumps http4s to 0.19.0 (which is [still not final](https://github.com/http4s/http4s/pull/2153), I guess?)
* rebased on top of current master
* makes all tests green
* Bumps SBT to latest version
* Makes json4s a test-only dependency for swagger

#### TODO

- [x] Fix `HListToFuncSpec`
- [x] Fix docstring in `AuthedContext`
- [x] Rename few more "service" to "routes"
- [x] Add a comment in `AuthedContext` that middleware is mandatory
- [x] Should we rename `RhoService` to `RhoRoutes`?
- [ ] Bump to final 0.20.0 when available

### DIFF
--- a/core/src/main/scala/org/http4s/rho/AuthedContext.scala
+++ b/core/src/main/scala/org/http4s/rho/AuthedContext.scala
@@ -10,6 +10,8 @@ import org.http4s.rho.bits.{FailureResponseOps, SuccessResponse, TypedHeader}
 
 /** The [[AuthedContext]] provides a convenient way to define a RhoRoutes
   * which works with http4s authentication middleware.
+  * Please note that `AuthMiddleware`-wrapping is mandatory, otherwise context
+  * doesn't take effect.
   * {{{
   *     case class User(name: String, id: UUID)
   *
@@ -48,7 +50,7 @@ class AuthedContext[F[_]: Monad, U] extends FailureResponseOps[F] {
     }
   }
 
-  /** Get the authInfo object from request */
+  /** Get the authInfo object from request if `AuthMiddleware` provided one */
   def getAuth(req: Request[F]): Option[U] =
     req.attributes.get(authKey)
 
--- a/core/src/main/scala/org/http4s/rho/AuthedContext.scala
+++ b/core/src/main/scala/org/http4s/rho/AuthedContext.scala
@@ -19,13 +19,13 @@ import org.http4s.rho.bits.{FailureResponseOps, SuccessResponse, TypedHeader}
   *
   *     object Auth extends AuthedContext[IO, User]
   *
-  *     object BobService extends RhoRoutes[IO] {
+  *     object BobRoutes extends RhoRoutes[IO] {
   *       GET +? param("foo", "bar") >>> Auth.auth |>> { (foo: String, user: User) =>
   *         Ok(s"Bob with id ${user.id}, foo $foo")
   *       }
   *     }
   *
-  *     val service = middleware.apply(Auth.toService(BobService.toRoutes()))
+  *     val service = middleware.apply(Auth.toService(BobRoutes.toRoutes()))
   * }}}
   *
   * @tparam U authInfo type for this service.
--- a/core/src/main/scala/org/http4s/rho/AuthedContext.scala
+++ b/core/src/main/scala/org/http4s/rho/AuthedContext.scala
@@ -49,7 +49,7 @@ class AuthedContext[F[_]: Monad, U] extends FailureResponseOps[F] {
     * @param service [[HttpService]] to convert
     * @return An `AuthedService` which can be mounted by http4s servers.
     */
-  def toService(service: HttpService[F]): AuthedService[U, F] = {
+  def toService(service: HttpRoutes[F]): AuthedService[U, F] = {
     type O[A] = OptionT[F, A]
 
     Kleisli[O, AuthedRequest[F, U], Response[F]] { (a: AuthedRequest[F, U]) =>
--- a/core/src/main/scala/org/http4s/rho/AuthedContext.scala
+++ b/core/src/main/scala/org/http4s/rho/AuthedContext.scala
@@ -44,16 +44,16 @@ class AuthedContext[F[_]: Monad, U] extends FailureResponseOps[F] {
   /* Attribute key to lookup authInfo in request attributeMap . */
   final private val authKey = AttributeKey[U]
 
-  /** Turn the [[HttpService]] into an `AuthedService`
+  /** Turn the [[HttpRoutes§]] into an `AuthedService`
     *
-    * @param service [[HttpService]] to convert
+    * @param routes [[HttpRoutes]] to convert
     * @return An `AuthedService` which can be mounted by http4s servers.
     */
-  def toService(service: HttpRoutes[F]): AuthedService[U, F] = {
+  def toService(routes: HttpRoutes[F]): AuthedService[U, F] = {
     type O[A] = OptionT[F, A]
 
-    Kleisli[O, AuthedRequest[F, U], Response[F]] { (a: AuthedRequest[F, U]) =>
-      service(a.req.withAttribute[U](authKey, a.authInfo))
+    Kleisli[O, AuthedRequest[F, U], Response[F]] { a: AuthedRequest[F, U] =>
+      routes(a.req.withAttribute[U](authKey, a.authInfo))
     }
   }
 
--- a/core/src/main/scala/org/http4s/rho/AuthedContext.scala
+++ b/core/src/main/scala/org/http4s/rho/AuthedContext.scala
@@ -8,7 +8,7 @@ import shapeless.{::, HNil}
 import org.http4s.rho.bits.{FailureResponseOps, SuccessResponse, TypedHeader}
 
 
-/** The [[AuthedContext]] provides a convenient way to define a RhoService
+/** The [[AuthedContext]] provides a convenient way to define a RhoRoutes
   * which works with http4s authentication middleware.
   * {{{
   *     case class User(name: String, id: UUID)
@@ -19,7 +19,7 @@ import org.http4s.rho.bits.{FailureResponseOps, SuccessResponse, TypedHeader}
   *
   *     object Auth extends AuthedContext[IO, User]
   *
-  *     object BobService extends RhoService[IO] {
+  *     object BobService extends RhoRoutes[IO] {
   *       GET +? param("foo", "bar") >>> Auth.auth |>> { (foo: String, user: User) =>
   *         Ok(s"Bob with id ${user.id}, foo $foo")
   *       }
--- a/core/src/main/scala/org/http4s/rho/AuthedContext.scala
+++ b/core/src/main/scala/org/http4s/rho/AuthedContext.scala
@@ -3,8 +3,9 @@ package rho
 
 import cats.Monad
 import cats.data.{Kleisli, OptionT}
-import org.http4s.rho.bits.{FailureResponseOps, SuccessResponse, TypedHeader}
 import shapeless.{::, HNil}
+
+import org.http4s.rho.bits.{FailureResponseOps, SuccessResponse, TypedHeader}
 
 
 /** The [[AuthedContext]] provides a convenient way to define a RhoService
@@ -12,29 +13,19 @@ import shapeless.{::, HNil}
   * {{{
   *     case class User(name: String, id: UUID)
   *
-  *     object Auth {
-  *       val authUser: Service[Request, User] = Kleisli({ _ =>
-  *         Task.now(User("Test User", UUID.randomUUID()))
-  *       })
-  *
-  *       val authenticated = AuthMiddleware(authUser)
+  *     val middleware = AuthMiddleware { req =>
+  *       OptionT(IO(User("Bob", UUID.randomUUID())))
   *     }
   *
-  *     object MyAuth extends AuthedContext[User]
+  *     object Auth extends AuthedContext[IO, User]
   *
-  *     object MyService extends RhoService {
-  *       import MyAuth._
-  *       GET +? param("foo", "bar") |>> { (req: Request, foo: String) =>
-  *         val user = getAuth(req)
-  *         if (user.name == "Test User") {
-  *           Ok(s"just root with parameter 'foo=\$foo'")
-  *         } else {
-  *           BadRequest("This should not have happened.")
-  *         }
+  *     object BobService extends RhoService[IO] {
+  *       GET +? param("foo", "bar") >>> Auth.auth |>> { (foo: String, user: User) =>
+  *         Ok(s"Bob with id ${user.id}, foo $foo")
   *       }
   *     }
   *
-  *     val service = Auth.authenticated(MyAuth.toService(MyService.toService(SwaggerSupport())))
+  *     val service = middleware.apply(Auth.toService(BobService.toRoutes()))
   * }}}
   *
   * @tparam U authInfo type for this service.
@@ -44,7 +35,7 @@ class AuthedContext[F[_]: Monad, U] extends FailureResponseOps[F] {
   /* Attribute key to lookup authInfo in request attributeMap . */
   final private val authKey = AttributeKey[U]
 
-  /** Turn the [[HttpRoutes§]] into an `AuthedService`
+  /** Turn the [[HttpRoutes]] into an `AuthedService`
     *
     * @param routes [[HttpRoutes]] to convert
     * @return An `AuthedService` which can be mounted by http4s servers.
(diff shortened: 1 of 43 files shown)


swagger/src/main/scala/org/http4s/rho/swagger/SwaggerModelsBuilder.scala
```diff
@@ -228,8 +228,8 @@ private[swagger] class SwaggerModelsBuilder(formats: SwaggerFormats) {
     Operation(
       tags        = pathStr.split("/").filterNot(_ == "").headOption.getOrElse("/") :: Nil,
       summary     = collectSummary(rr),
-      consumes    = rr.validMedia.toList.map(_.renderString),
-      produces    = rr.responseEncodings.toList.map(_.renderString),
+      consumes    = rr.validMedia.toList.map(_.toString),
+      produces    = rr.responseEncodings.toList.map(_.toString),
```
zarthross: I feel like the old `renderString` function was not the same as `toString` on a MediaRange. (Possibly not non an responseEncoding`)... `toString`'s results have a `MediaRange` and some parens...

zarthross: I'm looking at this diff in http4s to see the change: https://github.com/http4s/http4s/commit/f2ca17540cde57d0e5664ac5a9e9db37b92612c1

swagger/src/test/scala/org/http4s/rho/swagger/Arbitraries.scala
```diff
@@ -62,11 +62,11 @@ object Arbitraries {
...
-      `application/json`,
-      `audio/midi`,
-      `image/gif`,
-      `text/html`
-    ).map(_.renderString)
+      application.json,
+      audio.midi,
+      image.gif,
+      text.html
+    ).map(_.toString)
```
zarthross: Again, not sure `toString` should be the substitute for `renderString` on a `MediaRange`

examples/src/main/scala/com/http4s/rho/hal/plus/swagger/demo/Main.scala
```diff
@@ -18,16 +15,15 @@ object Main extends StreamApp[IO] {
...
+    val businessLayer = new UADetectorDatabase(new ResourceModuleXmlDataStore())
 
     val routes =
       new Routes(businessLayer)
 
     BlazeBuilder[IO]
       .mountService(routes.staticContent combineK routes.dynamicContent, "")
       .bindLocal(port)
-      .serve
+      .serve.compile.toList.map(_.head)
```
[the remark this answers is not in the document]
chuwy: Right, thanks. We don't want `toList` for sure. But I belive `compile.drain.as(ExitCode.Success)` is a preferred option for streaming applications inside `IOApp`.

swagger/src/test/scala/org/http4s/rho/swagger/Arbitraries.scala
chuwy: It seems that we have `Show` instance for `MediaType`. I believe it will be the correct substitution.

zarthross: I believe you are correct.

chuwy:
Thank you very much for the feedback, @zarthross. I'm still a little bit puzzled by `HListToFuncSpec` as it cannot resolver `EntityEncoder[IO, IO[String]]`, but probably I'm missing something super obvious as it is quite late in where I am - I'll try to fix it tomorrow.

chuwy:
Aha! So, the problem is that indeed `EntityDecoder[F, F[A]]` is missing since [this commit](https://github.com/http4s/http4s/commit/13527d6c320efd25800680a8fd841365dec7965e#diff-2f7b4e737e279d203ad95a9733e8cc94). Wondering what was the reasoning.

core/src/test/scala/org/http4s/rho/CodecRouterSpec.scala
```diff
@@ -14,7 +14,7 @@ class CodecRouterSpec extends Specification {
 
   "A CodecRouter in a RhoService" should {
 
-    val service = new RhoService[IO] {
+    val routes = new RhoService[IO] {
```
zarthross: This is still a `RhoService`, so i don't know that the rename is appropriate here.

core/src/test/scala/org/http4s/rho/ParamDefaultValueSpec.scala
```diff
@@ -6,340 +6,340 @@ import org.specs2.mutable.Specification
...
-    new String(service(r).value.unsafeRunSync().getOrElse(Response.notFound).body.compile.toVector.unsafeRunSync().foldLeft(Array[Byte]())(_ :+ _))
+  def body(routes: HttpRoutes[IO], r: Request[IO]): String =
+    new String(routes(r).value.unsafeRunSync().getOrElse(Response.notFound).body.compile.toVector.unsafeRunSync().foldLeft(Array[Byte]())(_ :+ _))
 
   def requestGet(s: String, h: Header*): Request[IO] =
     Request(bits.MethodAliases.GET, Uri.fromString(s).right.getOrElse(sys.error("Failed.")), headers = Headers(h: _*))
 
   "GET /test1" should {
-    val service = new RhoService[IO] {
+    val routes = new RhoService[IO] {
```
zarthross: This is still a `RhoService`, so i don't know that the rename is appropriate here.

core/src/test/scala/org/http4s/rho/ParamDefaultValueSpec.scala
```diff
@@ -6,340 +6,340 @@ import org.specs2.mutable.Specification
...
     }
     "map parameter without value" in {
-      body(service, requestGet("/test1?param1")) must be equalTo "Value of query parameter 'param1' missing"
+      body(routes, requestGet("/test1?param1")) must be equalTo "Value of query parameter 'param1' missing"
     }
   }
 
   "GET /test2" should {
-    val service = new RhoService[IO] {
+    val routes = new RhoService[IO] {
```
zarthross: This is still a `RhoService`, so i don't know that the rename is appropriate here.

core/src/test/scala/org/http4s/rho/ParamDefaultValueSpec.scala
```diff
@@ -6,340 +6,340 @@ import org.specs2.mutable.Specification
...
     }
     "map parameter without value" in {
-      body(service, requestGet("/test2?param1")) must be equalTo default
+      body(routes, requestGet("/test2?param1")) must be equalTo default
     }
   }
 
   "GET /test3" should {
-    val service = new RhoService[IO] {
+    val routes = new RhoService[IO] {
```
zarthross: as above

core/src/test/scala/org/http4s/rho/ParamDefaultValueSpec.scala
```diff
@@ -6,340 +6,340 @@ import org.specs2.mutable.Specification
...
     }
     "map parameter without value" in {
-      body(service, requestGet("/test3?param1")) must be equalTo default
+      body(routes, requestGet("/test3?param1")) must be equalTo default
     }
   }
 
   "GET /test4" should {
-    val service = new RhoService[IO] {
+    val routes = new RhoService[IO] {
```
zarthross: as above

core/src/test/scala/org/http4s/rho/ParamDefaultValueSpec.scala
```diff
@@ -6,340 +6,340 @@ import org.specs2.mutable.Specification
...
     }
     "map parameter without value" in {
-      body(service, requestGet("/test4?param1")) must be equalTo "test4:"
+      body(routes, requestGet("/test4?param1")) must be equalTo "test4:"
     }
   }
 
   "GET /test5" should {
-    val service = new RhoService[IO] {
+    val routes = new RhoService[IO] {
```
zarthross: as above

core/src/test/scala/org/http4s/rho/ParamDefaultValueSpec.scala
```diff
@@ -6,340 +6,340 @@ import org.specs2.mutable.Specification
...
     }
     "map parameter without value" in {
-      body(service, requestGet("/test5?param1")) must be equalTo default
+      body(routes, requestGet("/test5?param1")) must be equalTo default
     }
   }
 
   "GET /test6" should {
-    val service = new RhoService[IO] {
+    val routes = new RhoService[IO] {
```
zarthross: as above

core/src/test/scala/org/http4s/rho/ParamDefaultValueSpec.scala
```diff
@@ -6,340 +6,340 @@ import org.specs2.mutable.Specification
...
     "map parameter without value" in {
-      body(service, requestGet("/test13?param1")) must be equalTo default
+      body(routes, requestGet("/test13?param1")) must be equalTo default
     }
   }
 
   "GET /test14" should {
 
-    val service = new RhoService[IO] {
+    val routes = new RhoService[IO] {
```
zarthross: Gave up marking the all...

core/src/test/scala/org/http4s/rho/CodecRouterSpec.scala
chuwy: But we're immediately converting it to `HttpRoutes[IO]` below, so I think all these are justified.

core/src/main/scala/org/http4s/rho/CompileService.scala
```diff
@@ -42,7 +42,7 @@ object CompileService {
     * @param filter [[RhoMiddleware]] to apply to the routes.
     * @return An `HttpService`
     */
-  def foldServices[F[_]: Monad](routes: Seq[RhoRoute.Tpe[F]], filter: RhoMiddleware[F]): HttpService[F] = {
+  def foldServices[F[_]: Monad](routes: Seq[RhoRoute.Tpe[F]], filter: RhoMiddleware[F]): HttpRoutes[F] = {
```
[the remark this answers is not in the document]
zarthross: Hmm... TBH, I think this function needs a bit more than a rename, but it might be out of the scope of this PR.  So, yes i think you should rename it to `foldRoutes`.  

If you have time, I would say remove the `filter` parameter and have the [`ServiceBuilder.toService`](https://github.com/http4s/rho/blob/v0.18.0/core/src/main/scala/org/http4s/rho/ServiceBuilder.scala#L18) function which calls `foldServices` run the middleware on the routes before passing them in.  This reduces the complexity of this function, and makes the name make more sense.

core/src/test/scala/org/http4s/rho/CodecRouterSpec.scala
zarthross: ¯\_(ツ)_/¯ 
Not really that big of deal.  Not a deal breaker for me either way.

zarthross: (Removed other similar comments for brevity.)

rossabaker:
Yeah, sorry about 0.19.0 not being final.  I messed up the tagging.

Would you like to do a milestone release on this so people can try it?

zarthross:
I think there is a few more changes @chuwy would like to make before we merge/release, but once he finishes those check boxes, I would like to get this released.  I have some projects I would like to get a head start updating before the 0.19.1 release.

chuwy:
@rossabaker, @zarthross I made some more "service -> routes" renamings to be consistent with core and [updated AuthedContext](https://github.com/http4s/rho/pull/253/commits/ec2ba9e5e90656cd58d9aa5821ffec757141e26d) (I believe unsafe `getAuth` was a bug - I stumbled upon it when first used rho). It would be nice to get M1 when you have time.

Last question I have is whether we should rename `RhoService` to `RhoRoutes`. I vote for yes.

core/src/main/scala/org/http4s/rho/ServiceBuilder.scala
```diff
@@ -7,15 +7,15 @@ import shapeless.HList
...
 /** CompileService which accumulates routes and can build a `HttpService` */
-final class ServiceBuilder[F[_]: Monad] private(internalRoutes: VectorBuilder[RhoRoute.Tpe[F]]) extends CompileService[F, RhoRoute.Tpe[F]] {
+final class ServiceBuilder[F[_]: Monad] private(internalRoutes: VectorBuilder[RhoRoute.Tpe[F]]) extends CompileRoutes[F, RhoRoute.Tpe[F]] {
 
   /** Turn the accumulated routes into an `HttpService`
     *
     * @param filter [[RhoMiddleware]] to apply to the collection of routes.
     * @return An `HttpService` which can be mounted by http4s servers.
     */
   def toService(filter: RhoMiddleware[F] = identity): HttpRoutes[F] =
```
zarthross: `ServiceBuilder` and its `toService` function also need a rename.

core/src/main/scala/org/http4s/rho/AuthedContext.scala
```diff
@@ -57,13 +48,13 @@ class AuthedContext[F[_]: Monad, U] extends FailureResponseOps[F] {
     }
   }
 
-  /* Get the authInfo object from request. */
-  def getAuth(req: Request[F]): U = {
-    req.attributes.get[U](authKey).get
-  }
+  /** Get the authInfo object from request */
+  def getAuth(req: Request[F]): Option[U] =
+    req.attributes.get(authKey)
```
zarthross: Good catch, `get` should almost never be called on an `Option`, however, there is a more fundamental problem here that is beyond the scope of this PR to fix.   You should not be able to call `getAuth` if your `RhoService` hasn't been wrapped in a `AuthMiddleware` but the current design allows this.  So, for now, I propose we just make the error message you get for calling `getAuth` better, and file an issue to correct our auth framework so it doesn't allow this.

Thinking replacing `get` with `getOrElse(throw new RuntimeException("Misconfiguration: RhoService called getAuth but has not had AuthMiddleware applied"))`.  Certainly open to other exception messages, but looking for something more descriptive than `get` would throw alone.

core/src/main/scala/org/http4s/rho/ServiceBuilder.scala
chuwy: Agree, but I think we should handle this along with `RhoService`

core/src/test/scala/org/http4s/rho/AuthedContextSpec.scala
```diff
@@ -26,13 +26,22 @@ object MyAuth extends AuthedContext[IO, User]
...
       BadRequest("This should not have happened.")
     }
   }
+
+  GET / "public" / 'place |>> { path: String => Ok(s"not authenticated at $path") }
+
+  GET / "private" / 'place |>> { (req: Request[IO], path: String) =>
+    getAuth(req) match {
+      case Some(user) => Ok(s"${user.name} at $path")
+      case None => Forbidden(s"not authenticated at $path")
```
zarthross: Unfortunately, if `getAuth` returns `None` it doesn't mean the user was not authenticated, it means the `AuthMiddleware` wasn't applied.

core/src/main/scala/org/http4s/rho/ServiceBuilder.scala
zarthross: Yeah, I agree with the `RhoService` rename.  It makes sense in context with what http4s is trying to do.

core/src/main/scala/org/http4s/rho/AuthedContext.scala
chuwy: Raised it [here](https://github.com/http4s/rho/issues/256). For me `None` makes much more sense than `RuntimeException` because design of `AuthMiddleware` implies that it can return `None`.

core/src/test/scala/org/http4s/rho/AuthedContextSpec.scala
chuwy: Our `authenticated` middleware always returns `Some`, but in case where it returns `None` (which is still valid) - it will be indistinguishable from not applied middleware. So I have we have plenty of slightly broken APIs, but this one from my POV is less broken. We just need to mention in comments that `AuthMiddleware` always should be applied. Until we come up with best strategy.

core/src/main/scala/org/http4s/rho/ServiceBuilder.scala
chuwy: My worry is similar: we're getting too many "route" in different places that clash with each other and core's. We're ending up with `RouteFoo.toRoutes(routes).asRoutes` roughly speaking. But still I believe this is a right choice.

zarthross:
@chuwy How many of those checkboxes do you want to complete before this gets merged in?

chuwy:
@zarthross I think two at least. Should be done in next couple of days.

zarthross:
I think it would probably be wise to go through with the `RhoService` to `RhoRoutes` rename to follow suite with the Http4s rename.

chuwy:
Sure. Totally agree. I meant "all except Bump to final 0.19.1", so this renaming will be included for sure.

chuwy:
Also, I assume you're fine with `Option` in `AuthedContext`. I believe that it is a not an ideal solution, but still better than runtime exception.

project/Dependencies.scala
```diff
@@ -2,7 +2,7 @@ import sbt._
 import Keys._
 
 object Dependencies {
-  lazy val http4sVersion = "0.18.19"
+  lazy val http4sVersion = "0.19.0"
```
ChristopherDavenport: The currently version is `0.19.0-M4`, 0.19.0 was mistagged and should be avoided if possible.

chuwy:
Hey @zarthross, sorry for delay. This has been updated to http4s 0.20.0-M1. I also believe I addressed all your feedback and everything I wanted to change.

I assume we're not pursuing version-parity with http4s, so I left it as 0.19, not 0.20.

If everything is correct, it should be ready for M1.

rossabaker:
Version parity has always been coincidental, but maybe it's nice since it has recently been in lockstep.  I guess it might get more complicated when both projects hit 1.0. I don't have a strong opinion there.

PR #254, opened 2018-10-06
Documentation improvements

chuwy:
Snippets from current `Rho.md` do not compile and also do not describe all used constructs. I'm not sure my version is entirely correct and follow existing practices, but would save me quite some time when I first tried Rho.


zarthross:
'tender loving care', basically in this case... More work.

Issue #257, opened 2018-10-09
Optional CSV parameters rendered incorrectly in the Swagger


carwynellis:
Apologies if this issue has been reported before but I couldn't find any issues open or closed covering  this.

We have an optional CSV field on our API e.g.

`param[Option[Seq[String]]]("ids")`

However the parameter in the swagger is rendered as 

 ```json
{
  "name" : "foo",
  "in" : "query",
  "required" : false,
  "type" : "array",
  "items" : {
    "type" : "List"
  }
},
```

instead of 

```json
{
  "name" : "foo",
  "in" : "query",
  "required" : false,
  "type" : "array",
  "items" : {
    "type" : "string"
  }
},
```

Note that we need this parameter to be optional in the swagger e.g. `"required": false`.

If the parameter is defined as `param[Seq[String]]("ids")` then the swagger definition shows the correct type, e.g. array with items of type string, but the required property is then true.

The issue seems to be that during the swagger generation the type traversal of the `Option[Seq[String]]` stops at the `Seq` and doesn't continue to identify the actual type of the items.

RafalSumislawski:
@zarthross I think this issue can be closed.

PR #259, opened 2018-10-09
 Fix issue #257

zarthross:
Corrects issue #257 where an `Option` of some container will register as an array of that container instead of a not required container.

### DIFF
--- a/swagger/src/main/scala/org/http4s/rho/swagger/SwaggerModelsBuilder.scala
+++ b/swagger/src/main/scala/org/http4s/rho/swagger/SwaggerModelsBuilder.scala
@@ -78,7 +78,7 @@ private[swagger] class SwaggerModelsBuilder(formats: SwaggerFormats) {
             case _ : TypeBuilder.DataType.ComplexDataType =>
               tpe :: go(x::xs)
             case TypeBuilder.DataType.ContainerDataType(_, Some(_: TypeBuilder.DataType.ComplexDataType), _) =>
-              q.m.tpe.typeArgs.head :: go(x::xs)
+              q.m.tpe.dealias.typeArgs.head :: go(x::xs)
             case _ => go(x::xs)
           }
 
@@ -346,7 +346,7 @@ private[swagger] class SwaggerModelsBuilder(formats: SwaggerFormats) {
   def mkQueryParam[F[_]](rule: QueryMetaData[F, _]): Parameter = {
     val required = !(rule.m.tpe.isOption || rule.default.isDefined)
 
-    val tpe = if(rule.m.tpe.isOption) rule.m.tpe.typeArgs.head else rule.m.tpe
+    val tpe = if(rule.m.tpe.isOption) rule.m.tpe.dealias.typeArgs.head else rule.m.tpe
     TypeBuilder.DataType(tpe) match {
       case TypeBuilder.DataType.ComplexDataType(nm, _) =>
         QueryParameter(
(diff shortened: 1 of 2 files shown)


swagger/src/main/scala/org/http4s/rho/swagger/SwaggerModelsBuilder.scala
```diff
@@ -346,7 +346,8 @@ private[swagger] class SwaggerModelsBuilder(formats: SwaggerFormats) {
   def mkQueryParam[F[_]](rule: QueryMetaData[F, _]): Parameter = {
     val required = !(rule.m.tpe.isOption || rule.default.isDefined)
 
-    TypeBuilder.DataType(rule.m.tpe) match {
+    val tpe = if(rule.m.tpe.isOption) rule.m.tpe.typeArgs.head else rule.m.tpe
```
chuwy: Isn't `typeArgs.head` unsafe is type is an alias like `type OptInt = Option[Int]`? `OptInt` has zero type args.

zarthross: I'll have to check.

chuwy: Maybe `dealias` will be useful here https://github.com/xeno-by/scala/blob/0f4e95574081bd9a945fb5b32d157a32af840cd3/src/reflect/scala/reflect/api/Types.scala#L244

zarthross: You are correct.  I'll write up a test case and do something else.  Good catch @chuwy.

zarthross: So, 
```scala
      type OpSeq = Option[Seq[String]]
      val ra = fooPath +? param[OpSeq]("name") |>> { (s: OpSeq) => "" }
```

Doesn't compile anyways:
```
[error] /home/zarthross/Projects/rho/swagger/src/test/scala/org/http4s/rho/swagger/SwaggerModelsBuilderSpec.scala:105:39: No TypeTag available for OpSeq
[error]       val ra = fooPath +? param[OpSeq]("name") |>> { (s: OpSeq) => "" }
```

chuwy: Hm. I managed to compile it and get `NoSuchElementException`, but admit it was tricky

```
object OtherModule {
  type OpSeq = Option[Int]
  implicit val tt: TypeTag[OpSeq]
}

object OurSpec {
  import OtherModule._

  // remaining specification; fails because of head
}
```

chuwy: And `dealias` helped indeed. I also noticed few other `tpe.head` in that module. Probably we want to fix them as well.

zarthross: well, this is interesting:
```
scala> object Test {
     | type OpSeq = Option[Seq[String]]
     | val t = typeTag[OpSeq]
     | }
defined object Test
```
vs 
```
scala> def test {
     | type OpSeq = Option[Seq[String]]
     | return typeTag[OpSeq]
     | }
<console>:17: error: No TypeTag available for OpSeq
       return typeTag[OpSeq]
```

zarthross: Alright, added a few additional tests around this, and added another dealias that should help

chuwy: I assume that inside `def` `type` is not really global and that somehow prevents compiler from inspecting it. Quite superficial assumption, though, but yeah, that's interesting.

PR #270, opened 2018-11-17
Reworked handling of sum types. Solution for #268

### LINKED ISSUE #268: Apperance of a sum type affects models of other unrelated types.
Given the following code:
```scala
package rho.test

import java.util.concurrent.TimeUnit

import cats.effect.IO
import io.circe.generic.semiauto.deriveEncoder
import io.circe.{Encoder, Printer}
import org.http4s._
import org.http4s.circe.CirceInstances
import org.http4s.rho.RhoService
import org.http4s.rho.bits._
import org.http4s.rho.swagger.SwaggerSupport
import org.http4s.server.blaze.BlazeBuilder

import scala.util.Random

object RhoTest2 extends App {

  case class Baz(a: String)
  case class Baz1(a: String)
  sealed trait FooBar
  case class Foo(f: String) extends FooBar
  case class Bar(b: String) extends FooBar

  implicit val bazEncoder: Encoder[Baz] = deriveEncoder
  implicit val baz1Encoder: Encoder[Baz1] = deriveEncoder
  implicit val fooBarEncoder: Encoder[FooBar] = deriveEncoder
  implicit val fooEncoder: Encoder[Foo] = deriveEncoder
  implicit val barEncoder: Encoder[Bar] = deriveEncoder

  val mainService = new RhoService[IO] {

    private val circeInstances = CirceInstances.withPrinter(Printer.spaces2)
    private implicit def defaultJsonEncoder[A](implicit encoder: Encoder[A]): EntityEncoder[IO, A] =
      circeInstances.jsonEncoderOf[IO, A](stringEncoder, implicitly, encoder)

    private implicit def ioMatcher[R](implicit r: ResultMatcher[IO, R]): ResultMatcher[IO, IO[R]] = ResultMatcher.fMatcher

    GET / "baz" |>> { () => Ok(Baz("baz")) }
    GET / "baz1" |>> { () => Ok(Baz1("baz")) }
    GET / "foobar" |>> { () => Ok((if (Random.nextBoolean()) Foo("foo") else Bar("bar")): FooBar) }
  }

  val http = BlazeBuilder[IO]
    .bindHttp(55555, "localhost")
    .mountService(mainService.toService(SwaggerSupport[IO].createRhoMiddleware()), "/")
    .start.unsafeRunSync()

  TimeUnit.SECONDS.sleep(100000)
}
```
one would expect that the models of `Baz` and `Baz1` should differ just in name and have no connection to the `FooBar`. Rho generates the following swagger.json:
```json
{
  "swagger" : "2.0",
  "info" : {
    "version" : "1.0.0",
    "title" : "My API"
  },
  "paths" : {
    "/baz" : {
      "get" : {
        "tags" : [ "baz" ],
        "operationId" : "getBaz",
        "produces" : [ "application/json" ],
        "responses" : {
          "200" : {
            "description" : "OK",
            "schema" : {
              "$ref" : "#/definitions/Baz"
            }
          }
        }
      }
    },
    "/baz1" : {
      "get" : {
        "tags" : [ "baz1" ],
        "operationId" : "getBaz1",
        "produces" : [ "application/json" ],
        "responses" : {
          "200" : {
            "description" : "OK",
            "schema" : {
              "$ref" : "#/definitions/Baz1"
            }
          }
        }
      }
    },
    "/foobar" : {
      "get" : {
        "tags" : [ "foobar" ],
        "operationId" : "getFoobar",
        "produces" : [ "application/json" ],
        "responses" : {
          "200" : {
            "description" : "OK",
            "schema" : {
              "$ref" : "#/definitions/FooBar"
            }
          }
        }
      }
    }
  },
  "definitions" : {
    "Baz1" : {
      "type" : "object",
      "required" : [ "a" ],
      "properties" : {
        "a" : {
          "type" : "string"
        }
      },
      "description" : "Baz1"
    },
    "Bar" : {
      "allOf" : [ {
        "$ref" : "#/definitions/FooBar"
      }, {
        "type" : "object",
        "required" : [ "b" ],
        "properties" : {
          "b" : {
            "type" : "string"
          }
        },
        "description" : "Bar"
      } ],
      "description" : "Bar"
    },
    "FooBar" : {
      "type" : "object",
      "required" : [ "type" ],
      "discriminator" : "type",
      "properties" : {
        "type" : {
          "type" : "string",
          "enum" : [ "Bar", "Foo" ]
        }
      },
      "description" : "FooBar"
    },
    "Foo" : {
      "allOf" : [ {
        "$ref" : "#/definitions/FooBar"
      }, {
        "type" : "object",
        "required" : [ "f" ],
        "properties" : {
          "f" : {
            "type" : "string"
          }
        },
        "description" : "Foo"
      } ],
      "description" : "Foo"
    },
    "Baz" : {
      "allOf" : [ {
        "$ref" : "#/definitions/FooBar"
      }, {
        "type" : "object",
        "required" : [ "a" ],
        "properties" : {
          "a" : {
            "type" : "string"
          }
        },
        "description" : "Baz"
      } ],
      "description" : "Baz"
    }
  }
}
```

The interesting part is the model of Baz:
```json
"Baz" : {
      "allOf" : [ {
        "$ref" : "#/definitions/FooBar"
      }, {
        "type" : "object",
        "required" : [ "a" ],
        "properties" : {
          "a" : {
            "type" : "string"
          }
        },
        "description" : "Baz"
      } ],
      "description" : "Baz"
    }
```
The model refers to `FooBar` as if `Baz` was a subtype of the `FooBar` sealed trait. But it isn't. 

I've tracked the issue to the line [92 in TypeBuilder](https://github.com/http4s/rho/blob/d05998fdbd2f15df0b18bdd7f30437db02c766f0/swagger/src/main/scala/org/http4s/rho/swagger/TypeBuilder.scala#L92 ), where `composedModel` is applied to all models known so far instead of just subclasses of the sum type being processed. `Baz1` remained unaffected because it happened to be processed after `FooBar`. The code reproducing this issue may be unreliable as it depends on the order of traversal of a `Set[ResultInfo]`. 

Scala version: 2.12.7
Rho version: 0.18.0

RafalSumislawski:
This pull request:
* Fixes #268. Unrelated models are no longer accidentally modified.
* Changes behaviour in case only some subtypes of a sum type are in use. Now the whole sum type (the type and all it's children) are added to the model.
* Adds support for multi-level sum types. See the "Build a model for two-level sealed trait hierarchy" test case.
* Adds some tests, including some not directly related to sub types handling.
* Fixes some issues discovered in the process, namely: infinite recursion for `FooEither` and `BTree` types (see the test cases)
* cleans up/unifies the way `alreadyKnown` and `known` are used (mainly in order to fix the infinite recursion issues)

### DIFF
--- a/swagger/src/main/scala/org/http4s/rho/swagger/TypeBuilder.scala
+++ b/swagger/src/main/scala/org/http4s/rho/swagger/TypeBuilder.scala
@@ -7,7 +7,6 @@ import org.log4s.getLogger
 
 import scala.reflect.runtime.universe._
 import scala.util.control.NonFatal
-
 import cats.syntax.all._
 
 import scala.collection.immutable.ListSet
@@ -65,45 +64,33 @@ object TypeBuilder {
         case TypeRef(_, sym, _) if isObjectEnum(sym) =>
           Set.empty
 
-        case tpe@TypeRef(_, sym: Symbol, tpeArgs: List[Type]) if isCaseClass(sym) =>
-          val ownAndParentModel = sym.asClass.baseClasses.drop(1).find(isSumType) match {
-            case Some(parentSumType) =>
-              val parentModel = go(parentSumType.asType.toType, alreadyKnown, known + tpe)
+        case tpe@TypeRef(_, sym: Symbol, tpeArgs: List[Type]) if isCaseClass(sym) || isSumType(sym) =>
+          val symIsSumType = isSumType(sym)
+          val maybeParentSumType = sym.asClass.baseClasses.drop(1).find(isSumType)
 
+          val parentModel = maybeParentSumType.toSet[Symbol].flatMap(parentType => go(parentType.asType.toType, alreadyKnown, known + tpe))
+
+          val ownModel = maybeParentSumType match {
+            case Some(parentSumType) =>
               val parentType = parentSumType.asType.toType
               val parentRefModel = RefModel(parentType.fullName, parentType.simpleName, parentType.simpleName)
-              val ownModel = modelToSwagger(tpe, sfs).map(composedModel(parentRefModel)).toSet
-              parentModel ++ ownModel
-            case None =>
+              modelToSwagger(tpe, sfs).map(composedModel(parentRefModel)).toSet
+            case None if symIsSumType =>
+              modelToSwagger(tpe, sfs).map(addDiscriminator(sym)).toSet // only top level sum types get a discriminator
+            case None /*product type*/ =>
               modelToSwagger(tpe, sfs).toSet
           }
 
-          val fieldModels = sym.asClass.primaryConstructor.asMethod.paramLists.flatten.flatMap { paramsym =>
-            val paramType = paramsym.typeSignature.substituteTypes(sym.asClass.typeParams, tpeArgs)
-            go(paramType, alreadyKnown, known + tpe)
+          val childModels = if (symIsSumType) {
+            sym.asClass.knownDirectSubclasses.flatMap(childType => go(childType.asType.toType, alreadyKnown, known + tpe))
+          } else {
+            sym.asClass.primaryConstructor.asMethod.paramLists.flatten.flatMap { paramSym =>
+              val paramType = paramSym.typeSignature.substituteTypes(sym.asClass.typeParams, tpeArgs)
+              go(paramType, alreadyKnown, known + tpe)
+            }
           }
 
-          ownAndParentModel ++ fieldModels
-
-
-        case tpe@TypeRef(_, sym, _) if isSumType(sym) =>
-          // TODO promote methods on sealed trait from children to model
-          val ownAndParentModel = sym.asClass.baseClasses.drop(1).find(isSumType) match {
-            case Some(parentSumType) =>
-              val parentModel = go(parentSumType.asType.toType, alreadyKnown, known + tpe)
-
-              val parentType = parentSumType.asType.toType
-              val parentRefModel = RefModel(parentType.fullName, parentType.simpleName, parentType.simpleName)
-              val ownModel = modelToSwagger(tpe, sfs).map(composedModel(parentRefModel)).toSet
-              parentModel ++ ownModel
-            case None =>
-              modelToSwagger(tpe, sfs).map(addDiscriminator(sym)).toSet
-          }
-
-          val childTypeModels = sym.asClass.knownDirectSubclasses.flatMap { childType =>
-            go(childType.asType.toType, alreadyKnown, known + tpe)
-          }
-          ownAndParentModel ++ childTypeModels
+          parentModel ++ ownModel ++ childModels
 
         case _ => Set.empty
       }
--- a/swagger/src/main/scala/org/http4s/rho/swagger/TypeBuilder.scala
+++ b/swagger/src/main/scala/org/http4s/rho/swagger/TypeBuilder.scala
@@ -26,6 +26,12 @@ object TypeBuilder {
     def go(t: Type, alreadyKnown: Set[Model], known: Set[Type]): Set[Model] =
       t.dealias match {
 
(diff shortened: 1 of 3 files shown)


swagger/src/test/scala/org/http4s/rho/swagger/TypeBuilderSpec.scala
```diff
@@ -309,35 +323,86 @@ class TypeBuilderSpec extends Specification {
...
     }
 
-    "Build an empty model for an AnyVal" in {
-      val ms = modelOf[FooVal]
-      ms must be empty
-    }
+    // fails because the implementation for AnyVals is just Seg.empty
+    "Treat AnyVals transparently" in {
+      modelOf[FooVal] must_== modelOf[Foo]
+    }.pendingUntilFixed("AnyVals")
```
zarthross: I agree its a problem that `AnyVals` are not treated transparently.  Do you intend to change this behavior?  I don't like the idea of just leaving these pending.  Do you mind filing an issue if you do not intend to change this as part of this PR?

RafalSumislawski: I didn't want to increase the scope of this PR, IMO it's a candidate for a separate one. I just didn't want it to be forgotten, so I wrote a test for it. I'll fill an issue, and if I'll find some time in next days, I'll submit a second PR.

swagger/src/test/scala/org/http4s/rho/swagger/SwaggerModelsBuilderSpec.scala
```diff
@@ -49,7 +49,7 @@ class SwaggerModelsBuilderSpec extends Specification {
   implicit def defaultCompiler: CompileRoutes[IO, RhoRoute.Tpe[IO]] =
     CompileRoutes.identityCompiler
 
-  sealed abstract class Renderable
+  trait Renderable
```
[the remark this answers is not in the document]
RafalSumislawski: `Renderable` was `sealed`  so the `ModelBuilder` after my changes was treating it as a sum type, and so the generated model became more complex. The tests using ModelA, etc.. where not really about sum types and the `Renderable` was not really intended to be a sum type. It's here just to be able to constraint the `T` in `renderableEncoder[F[_], T <: Renderable]` an unsealed trait / abstract class is enough for that purpose, though in real world use cases one would rather use a type class.

swagger/src/test/scala/org/http4s/rho/swagger/TypeBuilderSpec.scala
RafalSumislawski: #272

swagger/src/main/scala/org/http4s/rho/swagger/TypeBuilder.scala
```diff
@@ -309,4 +307,13 @@ object TypeBuilder {
     private[this] def isCollection(t: Type): Boolean =
       t <:< typeOf[collection.Traversable[_]] || t <:< typeOf[java.util.Collection[_]]
   }
+
+  implicit class WrappedType(val t: Type){
```
zarthross: Mark this private so it won't leave this file... I usually try to avoid implicit conversions like this, but I understand the utility here so its fine as long as its localized.

swagger/src/main/scala/org/http4s/rho/swagger/TypeBuilder.scala
```diff
@@ -309,4 +307,13 @@ object TypeBuilder {
...
+    override def equals(obj: Any): Boolean = obj match{
+      case wt2: WrappedType => t =:= wt2.t
+      case _ => false
+    }
+  }
+
+  type TypeSet = ListSet[WrappedType]
```
zarthross: Also mark this private so it doesn't escape the scope.

RafalSumislawski:
@zarthross I've made the changes you've requested. 
Regarding the versions, I can wait for 0.19. There's no hurry.

Issue #262, opened 2018-10-14
Add HTTP Status Codes 103, 421, and 425


zarthross:
https://github.com/http4s/http4s/pull/2170

PR #271, opened 2018-11-18
Issue/2170 new status codes

zarthross:
Updated Location based status codes to have bodies.
Refactored Result graph so its easier to read.
Corrected a few misalignment's.

Fixes issue #262


core/src/main/scala/org/http4s/rho/Result.scala
```diff
@@ -7,79 +7,83 @@ import org.http4s.headers.`Content-Type`
...
+/* 207 */ +MULTISTATUS,
+/* 208 */ +ALREADYREPORTED,
+/* 226 */ +IMUSED,
+
+/* 300 */ +MULTIPLECHOICES,
+/* 301 */ +MOVEDPERMANENTLY,
+/* 302 */ +FOUND,
+/* 303 */ +SEEOTHER,
+/* 304 */ +NOTMODIFIED,
+/* 305 */ /* USEPROXY is deprecated https://tools.ietf.org/html/rfc7231#section-6.4.5*/
```
rossabaker: Does it hurt anything to continue support for it?  In http4s-core, I've been amazed by the things long presumed dead that people need to do to interface with some other product.  Though it's probably less of a problem serverside than clientside.  I trust your call either way.

zarthross: Its not that too hard to put it back in.  It was already partially removed which is why I just finished the job.  If Http4s supports it, Rho probably should support it too, so I'll put that back in and merge.

zarthross: Seems Http4s has taken this approach: ` // Note: UseProxy is deprecated in RFC7231, so we will not ease its creation here.`  So I'll do the same.  Any user can still just return a response with that status code, it won't get the metadata, and it won't be easy, but they can if they want to.

PR #279, opened 2019-01-01
WIP: New Authentication for Rho v2

zarthross:
This still needs lots of tests, and improvements with usages.

- [X] Ability to merge authed and unauthed services swagger docs.
- [ ] Lots more Test.
- [ ] Ability to add some auth information to swagger.
- [ ] Syntactic sugar for rejecting a request for some users on a per route basis.
- [ ] Syntactic sugar for `AuthedRhoRoutes[Kleisli[F, User, ?], User]` to `AuthedRhoRoutes[F, User]` 

This is different than #265 because of with this method a user doesn't have to deal with different [F] contexts forcing the generation of different swagger docs, or having to combine the swagger docs manually.

### DIFF
--- a/core/src/main/scala/org/http4s/rho/bits/RequestLike.scala
+++ b/core/src/main/scala/org/http4s/rho/bits/RequestLike.scala
@@ -1,0 +1,62 @@
+package org.http4s.rho.bits
+
+import org.http4s.{AttributeMap, AuthedRequest, EntityBody, Headers, HttpVersion, Method, Request, Uri}
+
+case class RequestLikeOps[R[_[_]], F[_]](r: R[F]) extends AnyVal {
+  def toRequest(implicit rl: RequestLike[R]): Request[F] = rl.toRequest(r)
+  def method(implicit rl: RequestLike[R]): Method = rl.method(r)
+  def uri(implicit rl: RequestLike[R]): Uri = rl.uri(r)
+  def httpVersion(implicit rl: RequestLike[R]): HttpVersion = rl.httpVersion(r)
+  def headers(implicit rl: RequestLike[R]): Headers = rl.headers(r)
+  def body(implicit rl: RequestLike[R]): EntityBody[F] = rl.body(r)
+  def attributes(implicit rl: RequestLike[R]): AttributeMap = rl.attributes(r)
+  def pathInfo(implicit rl: RequestLike[R]): String = rl.pathInfo(r)
+}
+
+trait RequestLike[R[_[_]]] {
+  def toRequest[F[_]](r: R[F]): Request[F]
+
+  def method[F[_]](r: R[F]): Method
+  def uri[F[_]](r: R[F]): Uri
+  def httpVersion[F[_]](r: R[F]): HttpVersion
+  def headers[F[_]](r: R[F]): Headers
+  def body[F[_]](r: R[F]): EntityBody[F]
+  def attributes[F[_]](r: R[F]): AttributeMap
+
+  def pathInfo[F[_]](r: R[F]): String
+}
+
+object RequestLike {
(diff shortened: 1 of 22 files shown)


core/src/main/scala/org/http4s/rho/bits/RequestLike.scala
```diff
@@ -0,0 +1,62 @@
...
+  def method(implicit rl: RequestLike[R]): Method = rl.method(r)
+  def uri(implicit rl: RequestLike[R]): Uri = rl.uri(r)
+  def httpVersion(implicit rl: RequestLike[R]): HttpVersion = rl.httpVersion(r)
+  def headers(implicit rl: RequestLike[R]): Headers = rl.headers(r)
+  def body(implicit rl: RequestLike[R]): EntityBody[F] = rl.body(r)
+  def attributes(implicit rl: RequestLike[R]): AttributeMap = rl.attributes(r)
+  def pathInfo(implicit rl: RequestLike[R]): String = rl.pathInfo(r)
+}
+
+trait RequestLike[R[_[_]]] {
```
rossabaker: This may be something that ties into @ChristopherDavenport's idea of a typeclass-based message.

core/src/main/scala/org/http4s/rho/bits/package.scala
```diff
@@ -0,0 +1,5 @@
+package org.http4s.rho
+
+package object bits {
```
rossabaker: This is a strange place for a public API.

zarthross: I wasn't intending for 'RequestLike' to be public.  Its only used internally right now to deal with the difference between a `Request` and `AuthedRequest`

zarthross:
I tried to keep the `RhoRoutes` parts basically the same, so the differences would mostly show up where they are converted to a `HttpRoutes`

I'll list the pro's and cons for both approaches here:

#265: `Kleisli` approach:
Pros:
  - It's relatively simple to implement

Cons:
 - The user will have to merge swagger docs for authed and not authed routes separate from conversion to `HttpRoutes`.
   - You can't just `and` together authed and un-authed routes, so you can't generate swagger from there.
   - We could create a `Monoid` (ish/like? thing) for swagger docs that lets you `|+|` combine the docs and let our users join the services separately, however they user would have to ensure the doc, and the services where composed in the same order since that kinda matters here....
  

#279: `AuthedRequest` approach:
Pros:
 - The user can merge Authed and Not Authed routes, and thus the swagger docs.
   - The user will need to provide the `AuthMiddleware` for a route before they can compose them with non-authed routes, but they can merge them now.

Cons:
 - Overly complex deep changes to the API

rossabaker:
I tend to prefer pros that favor the user over pros that favor the implementer, and that argues for #279 over #265.  I guess my worry about user pain would be increased difficulty reading the types or deciphering their compiler errors, but I think rho has always been about simplifying code/docs at the expense of complicated types.  This seems fully within the spirit of rho.

I haven't written a rho service in a long time, so don't weigh my opinion heavily, but I would vote for this.

zarthross:
I agree with `prefer pros that favor the user over pros that favor the implementer` fully.  
Seems I forgot to add a 'con' about the increase in complexity of the Rho Middleware for #279, since there are now 3 different types of Routes, there will be more difficulty in crafting those... though honestly beyond the swagger middleware, I'm not sure how much this is used in the wild....

rossabaker:
There was talk about deprecating HAL a while back and one person spoke up.  Maybe time to ask again.  Sadly, nothing brings users out of the shadows more than deleting a feature that they used.

zarthross:
Closing this in favor of #265 as its a much more simple/elegant change.

Issue #216, opened 2018-02-06
Implicit EntityEncoder for circe Encoder breaks compilation


DStranger:
Hi guys,

The following code doesn't compile with a `could not find implicit value for parameter hltf: org.http4s.rho.bits.HListToFunc[F,String :: shapeless.HNil,String => F[Test.this.Ok.T[String]]]`.
```scala
import cats.effect.Effect
import io.circe.Encoder
import org.http4s._
import org.http4s.rho.RhoService
import org.http4s.rho.swagger.SwaggerSyntax

trait Auto

class Test[F[_]](swagger: SwaggerSyntax[F])(implicit E: Effect[F]) extends RhoService[F] {
  import swagger._

  private implicit def autoEncoder[A <: Auto: Encoder]: EntityEncoder[F, A] = ???

  "This route allows your to post stuff" **
    POST / "post" ^ EntityDecoder.text[F] |>> { body: String =>
      Ok("you posted " + body)
  }
}
```

Removing `autoEncoder`, or removing the `io.circe.Encoder` constraint fixes the problem.

bryce-anderson:
Curious, since I believe the interesting part is that there isn't any need for the `Auto` entity encoder anyway. Do you need to remove the `io.circe.Encoder` import, or just the constraint in the `autoEncoder` method?

DStranger:
@bryce-anderson in this code there's indeed no need for `Auto`, but I've adapted it from some code that finds this construct useful (adapted from the swagger example).
Removing just the constraint works.

DStranger:
@bryce-anderson the problem resolves if I disable partial unification

bryce-anderson:
Interesting.

SystemFw:
that `EntityEncoder` should not be defined like that. I don't know if it's related to this specific problem or not, but might worth having a look at this:
https://github.com/http4s/http4s/issues/1648

DStranger:
@SystemFw yes, I don't think that's related, the error in http4s/http4s#1648 was caused by `jsonEncoderOf` requiring an instance of `EntityEncoder[F, String]`, here it's not the case.

SystemFw:
The puzzling thing here is the interaction with partial unification

Igosuki:
I have the same problem. circe.generic.auto works fine for EntityDecoders but not Encoders with Rho...

Igosuki:
I fixed it by hard encoding the import statements in the right order : 
  import swaggerSyntax._
  import io.circe.generic.auto._
[...]

PR #285, opened 2019-01-20
Fix issue #216

zarthross:
Removing the `Lazy` on the `ResultMatcher` (which is only barely recursive on situations of `F[ResultMatcher]` seems to fix the issue in #216 with circe `Encoder` causing a compilation issue.

Reproduction requires this dependency: 
`"io.circe" %% "circe-core" % "0.10.0"`

Minimum reproduction.
```
import cats.effect.Effect
import io.circe.Encoder
import org.http4s._
import org.http4s.rho.RhoRoutes

trait Auto

class Test[F[+_]: Effect] extends RhoRoutes[F] {
  implicit def autoEncoder[A <: Auto : Encoder]: EntityEncoder[F, A] = ???

  GET / "get" |>> ({ () => // Didn't work.  Used Lazy[ResultMatcher]
    Ok(())
  })

  GET / "get2" |>> ({ // Worked.  Does not use Lazy[ResultMatcher]
    Ok(())
  })
}
```

I'm still wondering how I can make a test, I haven't isolated what it is about the circe `Encoder` that causes the root issue.

### DIFF
--- a/core/src/main/scala/org/http4s/rho/bits/HListToFunc.scala
+++ b/core/src/main/scala/org/http4s/rho/bits/HListToFunc.scala
@@ -30,14 +30,14 @@ trait MatchersHListToFunc[F[_]] {
   }
 
   /** Converter for types `FunctionN` to an `HList` */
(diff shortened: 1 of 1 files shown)


danxmoran:
@zarthross is there a snapshot artifact of this branch which I could pull into my build?

zarthross:
No, but I'm going to go ahead and release this as part of the next
milestone, which will be either tonight or tomorrow.

PR #300, opened 2019-04-03
Fix Header capture 'default' semantics

zarthross:
The `capture*` family of functions had a `default` parameter that allowed a default response to be sent when a header was not present, but did not have a way to set a default value for a header that was not present, or to optionally capture a header (that is, use an `Option`).  

This PR renames the old `default` parameter to make more sense, fixes the broken swagger values that was being set for it, and adds ways to default, or optionally capture headers.

### DIFF
--- a/core/src/main/scala/org/http4s/rho/bits/Metadata.scala
+++ b/core/src/main/scala/org/http4s/rho/bits/Metadata.scala
@@ -25,4 +25,4 @@ trait SecurityScopesMetaData extends Metadata {
 case class QueryMetaData[F[_], T](name: String, description: Option[String], p: QueryParser[F, T], default: Option[T], m: TypeTag[T]) extends Metadata
 
 /** Metadata about a header rule */
-case class HeaderMetaData[T <: HeaderKey.Extractable](key: T, default: Boolean) extends Metadata
+case class HeaderMetaData[T <: HeaderKey.Extractable](key: T, isRequired: Boolean) extends Metadata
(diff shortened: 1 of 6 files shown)


core/src/main/scala/org/http4s/rho/bits/Metadata.scala
```diff
@@ -25,4 +25,4 @@ trait SecurityScopesMetaData extends Metadata {
 case class QueryMetaData[F[_], T](name: String, description: Option[String], p: QueryParser[F, T], default: Option[T], m: TypeTag[T]) extends Metadata
 
 /** Metadata about a header rule */
-case class HeaderMetaData[T <: HeaderKey.Extractable](key: T, default: Boolean) extends Metadata
+case class HeaderMetaData[T <: HeaderKey.Extractable](key: T, isRequired: Boolean) extends Metadata
```
rossabaker: It ate my comment...

This could lead to boolean blindness vs. something like `Required | Optional`.  But your approach has the virtue of being less breaking.  I'd probably leave it like this, but just a thought.

zarthross: Hopefully people don't use the metadata directly often if ever.  Do you know of some ADT/Enum with `Required/Optional` that i can plugin here?  I don't really feel like it would be worth the effort (at this time anyways) to add them if there isn't some ready made solution already a dependency.

rossabaker: No, nothing standard.  Roll with what's here, I'd say.

PR #302, opened 2019-04-04
Fix issue 301, POST with body and query string, with no header extractor

zarthross:
There was a missing link in the chain
PathBuilder -> QueryBuilder -> CodecRouter

What was available
PathBuilder -> QueryBuilder -> Router -> CodecRouter
PathBuilder -> Router -> CodecRouter
PathBuilder -> CodecRouter


zarthross:
It was already supported, you were just forced to also extract a header....

PR #307, opened 2019-04-06
WIP Scala 2.13.0-M5

zarthross:


zarthross:
Good call-out.   I was already considering making uses of Seq explicitly immutable.. so, i'll just go ahead and do that.

rossabaker:
Just a note on roadmap: We'll keep http4s-0.20 on M5, and target http4s-0.21 for the release of 2.13.0-final.  If we can get a milestone out on RC2 beforehand, we will.

RafalSumislawski:
Hi @zarthross 
I see there was no progress on this recently. Maybe I could help and pick it up where you left off?

RafalSumislawski:
or is the plan to merge #324 instead?

zarthross:
This branch, right now mainly has changes from `scala.collection.Seq` to `scala.collections.immutable.Seq` which I think it still something we want for the next version of Rho, but I don't think is in that other PR.  I think I'll open a separate PR for that change and close this one in favor of the other.

zarthross:
Scratch that, I forgot I already pulled those immutable seq fixes out of this PR.

PR #310, opened 2019-04-09
Custom extractor errors

zarthross:


core/src/main/scala/org/http4s/rho/RhoDslHeaderExtractors.scala
```diff
@@ -121,9 +121,19 @@ trait RhoDslHeaderExtractors[F[_]]
...
       } catch {
         case NonFatal(e) =>
-          logger.error(e)(s"""Failure during header capture: "${key.name}" ${h.fold("Undefined")(v => s"""= "${v.value}"""")}""")
-          error("Error processing request.")
+          FailureResponse.result(errorProcessingHeaderResponse(key, h, e))
       }
     }.withMetadata(HeaderMetaData(key, isRequired = isRequired))
+
+  protected def invalidHeaderResponse[H <: HeaderKey.Extractable](h: H#HeaderT)(implicit F: Monad[F]): F[BaseResult[F]] =
+    BadRequest(s"Invalid header: ${h.name} = ${h.value}").widen
```
rossabaker: It is a bit risky to reflect data from the response back in the request.  It's much harder to exploit with `text/plain` than other content types, but even this _can_ cause XSS issues.  http4s core does not do this by default for this reason, though it does make the errors a bit harder for legitimate clients.

zarthross: What's your suggestion then?  Just provide the header key rather than the header itself?
Or provide the header for users that want it, but start doing the better thing by default?

rossabaker: The header key is user input too, so that's not safer than the value.  Not rendering these is the difference between `sanitized` and `details` in the `MessageFailure` hierarchy of http4s-core.  Those failures are able to carry both, so users can recover the errors and reveal the level of detail they want.  I'm not sure if we can smuggle that through this `BadRequest` type or not.  An ideal design will let the user tune convenience vs. paranoia to their preferred taste, but the http4s-core defaults might be stricter than many people need.  (Our default paranoia was inherited from spray, and I worked in cybersecurity at the time, so we kept it.)

zarthross: Ok, I looked a bit more in the XSS stuff you mentioned, i think the header key is safe in this case, because the header keys here come from code, `exists(headerKey)`.  As for the value, yeah, a XSS reflection attack would not be ideal.  I agree with sanitizing it so the function only gets the key, and not the value to reduce users messing up.  Besides, if you get a bad request because of the header you should already know the value, it would be helpful to know which header messed up though.

zarthross:
@rossabaker I restricted the header errors so they don't reflect the values, let me know if this works for you.

rossabaker:
Looks good to me.

PR #324, opened 2019-06-20
Upgrade to scala 2.13 and http4s 0.21.0

Igosuki:


build.sbt
```diff
@@ -94,7 +91,7 @@ lazy val license = licenses in ThisBuild := Seq(
 
 lazy val buildSettings = publishing ++
   Seq(
-    scalaVersion := "2.12.8",
+    scalaVersion := "2.13.0",
     crossScalaVersions := Seq(scalaVersion.value, "2.11.12"),
```
lorandszakacs: `crossScalaVersions` also needs  `"2.12.8"`

build.sbt
```diff
@@ -79,9 +78,7 @@ lazy val compileFlags = Seq(
   "-language:higherKinds",
   "-language:existentials",
   "-language:implicitConversions",
-  "-Ywarn-unused",
-  "-Ypartial-unification",
-  "-Xfatal-warnings"
+  "-Ywarn-unused"
```
lorandszakacs: Flags can be configured based on scala version. Something like this:
```
scalacOptions ++= (CrossVersion.partialVersion(scalaVersion.value) match {
        case Some((2, 11)) => scala2_11Flags
        case Some((2, 12)) => scala2_12Flags
        case Some((2, 13)) => scala2_13Flags
        case _             => Seq.empty
```

Igosuki:
Yeah sorry I'll put out a compat folder like in circe :) coming back soon

Igosuki:
Doesn't compile with 2.12 because some partial unification is failing on CompilerRoutes, trying to figure out why...

build.sbt
```diff
@@ -95,8 +107,8 @@ lazy val license = licenses in ThisBuild := Seq(
 lazy val buildSettings = publishing ++
   Seq(
     scalaVersion := "2.12.8",
-    crossScalaVersions := Seq(scalaVersion.value, "2.11.12"),
-    scalacOptions ++= compileFlags,
+    crossScalaVersions := Seq(scalaVersion.value, "2.12.8", "2.11.12"),
```
lorandszakacs: you no longer have cross-compilation for version `2.13.0`, since the main `scalaVersion` is `2.12.8` again :smile:

Igosuki: Yeah don't worry I'm doing changes right now

Igosuki: It's back now, I need to find a way to make the part using MapView cross compile in rho-swagger (or refactor) and then I think it'll be over.

Igosuki:
I fixed most issues, although now in 2.12, HttpRoutes doesn't have a SemigroupK, which blocks the examples module from compiling...

Igosuki:
It looks like partial unification doesn't work properly :/

Igosuki:
Hmm, it was a typo... shameful

Igosuki:
I added placeholders that are missing from the current scala-collection-compat latest release and used in the demo code, I put them all in core, but maybe it will cause conflicts for other users down the line ?

build.sbt
```diff
@@ -73,22 +85,32 @@ lazy val `rho-examples` = project
...
   "-Xfatal-warnings"
 )
 
+def versionSpecificEnabledFlags(version: String) = (CrossVersion.partialVersion(version) match {
+  case Some((2, 13)) => Seq.empty[String]
+  case _ => Seq("-Ypartial-unification")
+})
+
+def versionSpecificDisabledFlags(version: String) = (CrossVersion.partialVersion(version) match {
+  case Some((2, 13)) => Seq("-Xfatal-warnings")
```
zarthross: What is the reason for disabling fatal warnings for 2.13 ?

lorandszakacs:
@Igosuki, anything blocking here? I could pick up the branch if you're too busy or something like that.

build.sbt
RafalSumislawski: There's a false-positive "unreachable code" warning in `StaticContentService`. Cause most likely one of these bugs: https://github.com/scala/bug/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+unreachable+2.13

Igosuki:
It was working fine on my laptop, but please go ahead. I hadn't had news in a while so I didn't touch it.

zarthross:
@Igosuki Thank you for the initial work on this.  I thought this was still WIP since it had failing builds so I didn't give it a full review.  I'll try and be more proactive in the future.  I'm closing this in favor of #342

PR #342, opened 2019-09-16
Upgrade to scala 2.13 and http4s 0.21.0 (second attempt)

RafalSumislawski:
The work on #324 is stuck since over a month, so I've picked up the work by @Igosuki and made some changes to make it mergable.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -122,7 +122,7 @@ lazy val license = licenses in ThisBuild := Seq(
 lazy val buildSettings = publishing ++
   Seq(
     scalaVersion := "2.13.0",
-    crossScalaVersions := Seq(scalaVersion.value, "2.12.8", "2.11.12"),
+    crossScalaVersions := Seq(scalaVersion.value, "2.12.8"),
     scalacOptions := compilerFlags ++ versionSpecificEnabledFlags(scalaVersion.value),
     resolvers += Resolver.sonatypeRepo("snapshots"),
     fork in run := true,
--- a/build.sbt
+++ b/build.sbt
@@ -85,8 +85,7 @@ lazy val `rho-examples` = project
       Seq(
         exampleDeps,
         libraryDependencies ++= Seq(logbackClassic, http4sXmlInstances),
-        dontPublish,
-        scalacOptions --= versionSpecificDisabledFlags(scalaVersion.value)
+        dontPublish
       ): _*)
   .dependsOn(`rho-swagger`, `rho-hal`)
 
@@ -104,11 +103,6 @@ lazy val compilerFlags = Seq(
 def versionSpecificEnabledFlags(version: String) = (CrossVersion.partialVersion(version) match {
   case Some((2, 13)) => Seq.empty[String]
   case _ => Seq("-Ypartial-unification")
-})
-
-def versionSpecificDisabledFlags(version: String) = (CrossVersion.partialVersion(version) match {
-  case Some((2, 13)) => Seq("-Xfatal-warnings")
-  case _ => Seq.empty
 })
 
 /* Don't publish setting */
--- a/build.sbt
+++ b/build.sbt
@@ -15,7 +15,9 @@ lazy val rho = project
 
 lazy val `rho-core` = project
   .in(file("core"))
-  .settings(buildSettings: _*)
+  .settings(buildSettings ++ Seq(
+    libraryDependencies ++= Seq("org.scala-lang.modules" %% "scala-collection-compat" % "2.0.0")
+  ): _*)
 
 lazy val `rho-hal` = project
   .in(file("hal"))
@@ -94,7 +96,6 @@ def versionSpecificDisabledFlags(version: String) = (CrossVersion.partialVersion
   case _ => Seq.empty
 })
 
-
 /* Don't publish setting */
 lazy val dontPublish = packagedArtifacts := Map.empty
 
@@ -105,7 +106,7 @@ lazy val license = licenses in ThisBuild := Seq(
 
 lazy val buildSettings = publishing ++
   Seq(
-    scalaVersion := "2.13.0",
+    scalaVersion := "2.12.8",
     crossScalaVersions := Seq(scalaVersion.value, "2.12.8", "2.11.12"),
     scalacOptions := compilerFlags ++ versionSpecificEnabledFlags(scalaVersion.value),
     resolvers += Resolver.sonatypeRepo("snapshots"),
--- a/build.sbt
+++ b/build.sbt
@@ -98,7 +98,7 @@ lazy val compilerFlags = Seq(
 
 def versionSpecificEnabledFlags(version: String) = (CrossVersion.partialVersion(version) match {
   case Some((2, 13)) => Seq.empty[String]
-  case _ => Seq("Ypartial-unification")
+  case _ => Seq("-Ypartial-unification")
 })
 
 def versionSpecificDisabledFlags(version: String) = (CrossVersion.partialVersion(version) match {
--- a/build.sbt
+++ b/build.sbt
@@ -42,6 +42,7 @@ lazy val docs = project
       version.value,
       apiVersion.value
     ),
+    scalacOptions in (ScalaUnidoc, unidoc) ++= versionSpecificEnabledFlags(scalaVersion.value),
     unidocProjectFilter in (ScalaUnidoc, unidoc) := inProjects(
       `rho-core`,
       `rho-hal`,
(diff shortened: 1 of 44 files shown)


core/src/test/scala/org/http4s/rho/RhoRoutesSpec.scala
```diff
@@ -369,7 +369,7 @@ class RhoRoutesSpec extends Specification with RequestRunner {
         GET / "bar" |>> "bar"
       }
 
-      val routes2: HttpRoutes[IO] = "foo" /: routes1 toRoutes()
+      val routes2: HttpRoutes[IO] = "foo" /: routes1 toRoutes(i => i) // No idea why identity won't compile
```
zarthross: I think we need to look into why this won't compile a bit further before we commit this.  I worry that will cause users issues.

version.sbt
```diff
@@ -1,2 +1,2 @@
-version       in ThisBuild := "0.19.1-SNAPSHOT"
+version       in ThisBuild := "0.21.0-M1"
```
zarthross: Rho has largely, by change, had the same version number as http4s, but nothing strickly ties us.  Its actually convenient to not be have the same number because we can release breaking changes out of cycle with Http4s.  I think I would rather the next version be 0.20... unless people think the different is to confusing to end users.

build.sbt
```diff
@@ -73,22 +85,32 @@ lazy val `rho-examples` = project
...
   "-Xfatal-warnings"
 )
 
+def versionSpecificEnabledFlags(version: String) = (CrossVersion.partialVersion(version) match {
+  case Some((2, 13)) => Seq.empty[String]
+  case _ => Seq("-Ypartial-unification")
+})
+
+def versionSpecificDisabledFlags(version: String) = (CrossVersion.partialVersion(version) match {
+  case Some((2, 13)) => Seq("-Xfatal-warnings")
```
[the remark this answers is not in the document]
RafalSumislawski: The whole fuss is about scalac 2.13 incorrectly issuing a warning in the pattern matching in `StaticContentService.routes`, so may proposition (https://github.com/http4s/rho/pull/342/commits/668da3a0bbd3d9b8617d8fb84a36f2a595669837) is to use RhoRoutes DSL instead.

core/src/test/scala/org/http4s/rho/RhoRoutesSpec.scala
RafalSumislawski: For some reason Scala 2.13 interprets `something toRoutes()` as a attempt to pass `()` to the `toRoutes`. In other words it desugars to `something.toRoutes(())`. `something.toRequest()` works as expected. So I propose using `("foo" /: routes1).toRoutes()`.

core/src/test/scala/ApiExamples.scala
```diff
@@ -39,8 +39,8 @@ class ApiExamples extends Specification {
         GET / "helloworldnumber" / pathVar[Int] / "foo" |>> { i: Int =>
           Ok(s"Received $i")
         }
-        // the symbol 'world just says 'capture a String' with variable name "world"
-        GET / "helloworldstring" / 'world / "foo" |>> { i: String =>
+        // the symbol Symbol("world") just says 'capture a String' with variable name "world"
+        GET / "helloworldstring" / Symbol("world") / "foo" |>> { i: String =>
```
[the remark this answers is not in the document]
RafalSumislawski: What would you say about `pv"world"` (`pv` referring to `pathVar`) instead?

build.sbt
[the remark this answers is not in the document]
zarthross: That works for me.

core/src/test/scala/org/http4s/rho/RhoRoutesSpec.scala
zarthross: I'm fine with adding the parens.

core/src/test/scala/ApiExamples.scala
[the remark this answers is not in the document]
zarthross: Like it.

core/src/main/scala/org/http4s/rho/RhoDslPathExtractors.scala
```diff
@@ -34,3 +42,13 @@ trait RhoDslPathExtractors[F[_]] {
     TypedPath(PathCapture[F](id, Some(description), parser, stringTag))
 
 }
+
+object RhoDslPathExtractors {
+
+  private val stringTag = implicitly[TypeTag[String]]
+
+  class PathCaptureStringContext[F[_]](val sc: StringContext) extends AnyVal {
+    def pv(args: Any*): TypedPath[F, String :: HNil] =
```
zarthross: So, I'm not sure about allowing args to the interpolater.

Are we ok with path names like this? 

```
val foo = "Some string from somewhere"

GET / "helloworldstring" / pv"world-$foo"

``` 

If this is desirable I'm OK with it, by I feel like we should restrict it to just other strings instead of `Any`?

examples/src/main/scala/com/http4s/rho/swagger/demo/Main.scala
```diff
@@ -19,10 +23,13 @@ object Main extends IOApp {
   logger.info(s"Starting Swagger example on '$port'")
 
   def run(args: List[String]): IO[ExitCode] = {
+    val blockingExecutor = Executors.newCachedThreadPool()
+    implicit val blocker: Blocker = Blocker.liftExecutionContext(ExecutionContext.fromExecutor(blockingExecutor))
```
zarthross: I know this is just the example project, but maybe we should set a good example by wrapping the execution context into a Resource that properly closes the execution context when we are done using it?

examples/src/main/scala/com/http4s/rho/swagger/demo/MyRoutes.scala
```diff
@@ -47,7 +47,7 @@ abstract class MyRoutes[F[+_] : Effect](swaggerSyntax: SwaggerSyntax[F])
     HEAD / "hello" |>> { Ok("Hello head!") }
 
   "Generates some JSON data from a route param, and a query Int" **
-    GET / "result" / Symbol("foo") +? param[Int]("id") |>> { (name: String, id: Int) => Ok(JsonResult(name, id)) }
+    GET / "result" / pv"foo" +? param[Int]("id") |>> { (name: String, id: Int) => Ok(JsonResult(name, id)) }
```
zarthross: :+1:

examples/src/main/scala/com/http4s/rho/swagger/demo/StaticContentService.scala
```diff
@@ -1,28 +1,27 @@
...
-import scala.concurrent.ExecutionContext.global
+import org.http4s.rho.RhoRoutes
+import org.http4s.{HttpRoutes, Request, Response, StaticFile}
 
 object StaticContentService {
   private val swaggerUiDir = "/swagger-ui"
 
-  def fetchResource(path: String, req: Request[IO])(implicit cs: ContextShift[IO]): IO[Response[IO]] = {
-    StaticFile.fromResource(path, Blocker.liftExecutionContext(global), Some(req)).getOrElseF(NotFound())
+  def fetchResource(path: String, req: Request[IO])(implicit cs: ContextShift[IO], blocker: Blocker): IO[Response[IO]] = {
```
zarthross: Again on this one, I know its an example, but lets set a good one, according to the [cats doc](https://typelevel.org/cats-effect/api/cats/effect/Blocker.html), instances of the `Blocker type`  ` should *not* be passed implicitly. `

version.sbt
lorandszakacs: @zarthross I think it's fine to go out of step with http4s as long as we put the http4s version used in bold, on the first line of each release description, and in other relevant parts of documentation.

core/src/main/scala/org/http4s/rho/RhoDslPathExtractors.scala
RafalSumislawski: For some reason I was convinced that to use that syntax I must define a method with `args: Any*`. I just learned that this isn't true. 

I would prefer to completely disable the interpolation.

examples/src/main/scala/com/http4s/rho/swagger/demo/Main.scala
aeons: On the JVM you can also create a blocker by `Blocker[F]`. 
https://github.com/typelevel/cats-effect/blob/ac3644d4afde1ee40c36f0449bdc4f202316889a/core/jvm/src/main/scala/cats/effect/internals/BlockerPlatform.scala#L29

RafalSumislawski: :+1: I support every change that makes the example more "production ready". Examples tend to end up being copied to real projects and run on production.

examples/src/main/scala/com/http4s/rho/swagger/demo/StaticContentService.scala
RafalSumislawski: Personally I'm not convinced that the drawbacks of passing it as implicit overweight the convenience of doing it. That being said, I agree that the examples should rather stay in line with the official recommendation than my personal preferences ;).

lorandszakacs:
@zarthross since there doesn't seem to be any additional feedback can you please merge this? And maybe even prepare a release please :D

zarthross:
@lorandszakacs. Done as requested.  It's out in mavem central now.

lorandszakacs:
@zarthross awesome. Thank you! 😄

PR #350, opened 2019-10-01
Add StringParser.map and StringParser.rmap

RafalSumislawski:
I would like to propose adding `map` and `rmap` functions to `StringParser`, in order to simplify defining new parsers on top of existing definitions. The idea is inspired by a similar solution in circe ([`io.circe.Decoder.map`](https://github.com/circe/circe/blob/master/modules/core/shared/src/main/scala/io/circe/Decoder.scala#L94) and [`io.circe.Decoder.emap`](https://github.com/circe/circe/blob/master/modules/core/shared/src/main/scala/io/circe/Decoder.scala#L335) )

### DIFF
--- a/core/src/main/scala/org/http4s/rho/bits/StringParser.scala
+++ b/core/src/main/scala/org/http4s/rho/bits/StringParser.scala
@@ -22,7 +22,13 @@ trait StringParser[F[_], T] extends ResponseGeneratorInstances[F] {
   /** TypeTag of the type T */
   def typeTag: Option[TypeTag[T]]
 
-  def invalidNumberFormat[A](n : String)(implicit F: Monad[F]): FailureResponse[F] = FailureResponse.pure[F] {
+  def map[U](f: T => U)(implicit ttu: TypeTag[U]): StringParser[F, U] =
+    new RMappedParser[F, T, U](this, (t: T) => SuccessResponse(f(t)), ttu)
+
+  def rmap[U](f: T => ResultResponse[F, U])(implicit ttu: TypeTag[U]): StringParser[F, U] =
+    new RMappedParser[F, T, U](this, f, ttu)
+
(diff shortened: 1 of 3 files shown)


RafalSumislawski:
I think I'm going to rework the `GET / "complex" +? param[Foo]("foo") & param[Seq[Bar]]("bar", Nil)` example to be a bit less complex and to showcase these new methods. I think it's bad as it is now, as a minimal correct request for that endpoint is `http://127.0.0.1:8080/complex?foo=%7B"k"%3A"a"%2C"v"%3A7%7D` (BTW it gets broken when you use it twice in Firefox, as firefox assumes curly braces don't have to be percent endoced, and parboiled has a different opinion on that). Passing jsons in query params shouldn't be something that's encouraged by the documentation/examples.

RafalSumislawski:
Regarding the swagger output. It depends on what one expects. Rho outputs illegal parameter types on a regular basis (see the `bar` in the `complex` example). This PR doesn't change anything in that area. Both `map` and `emap` capture the type tag of the new type, and that type tag is used to build the (sometimes illegal) swagger model. 

That being said, I think there is a possible way to improve the situation, though IMO outside of the scope of this PR. We could decouple the swagger model generation from the type of the value produced by the StringParser, and instead use a swagger type associated with one of the predefined StringParsers. For example:

given a `case class Foo(a: String, b: String)`, and a `param[Foo]` which currently would give an illegal swagger model. If we derived the `StringParser[Foo]` from a `StringParser.strParser[F]` through use of `map`/`rmap`, we would know that `string` is our best guess at finding a valid swagger type of the `param`.

RafalSumislawski:
I've adapted the example to use `map` and `rmap` (and not use things which don't work well like jsons in query params).

RafalSumislawski:
Regarding the swagger output. This PR doesn't change anything in that area. Both `map` and `emap` capture the type tag of the new type, and that type tag is used to build the swagger model. Just like with the old way of defining StringParsers, we often end up with documenting things as `string`, as that is the fallback in `DataType.fromType`. But I think it's a separate topic.

PR #358, opened 2019-10-25
Fix validation failures caused by Unit/Void/Nothing/Null

RafalSumislawski:
I've noticed that my swagger files fail validation when I have `case class Result[T](t: T msg: String)` and then use `Result[Unit]`. The problem is that rho emits a swagger schema with a property with `"type": "Unit"`. It isn't directly stated in swagger specification that `type` can't be an arbitrary string, but it turns out this requirement is inherited from JSON Schema.

This PR changes rho's behaviour when handling Unit/Void/Nothing/Null to:
- When building a model, emit a `RefProperty` referring to the definition of "Unit/Void/Nothing/Null" 
- Fall back to "string" where refs aren't allowed (query params, pathVars etc.)

### DIFF
--- a/swagger/src/test/scala/org/http4s/rho/swagger/SwaggerModelsBuilderSpec.scala
+++ b/swagger/src/test/scala/org/http4s/rho/swagger/SwaggerModelsBuilderSpec.scala
@@ -175,6 +175,16 @@ class SwaggerModelsBuilderSpec extends Specification {
         QueryParameter(`type` = None, name = "bar".some, items = Some(AbstractProperty(`type` = "string")), defaultValue = "".some, isArray = true)
       )
     }
+
+    "handle and action with query parameters of empty data types" in {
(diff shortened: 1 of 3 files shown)


swagger/src/test/scala/org/http4s/rho/swagger/SwaggerModelsBuilderSpec.scala
```diff
@@ -175,6 +175,16 @@ class SwaggerModelsBuilderSpec extends Specification {
         QueryParameter(`type` = None, name = "bar".some, items = Some(AbstractProperty(`type` = "string")), defaultValue = "".some, isArray = true)
       )
     }
+
+    "handle and action with query parameters of empty data types" in {
```
zarthross:
```suggestion
    "handle an action with query parameters of empty data types" in {
```

Issue #366, opened 2019-12-06
Provide example using the Swagger UI Webjar


etspaceman:
Hello there - 

I was going through your examples when setting up a project of my own, and I noticed that you are including the swagger-ui distribution directly in the project's resources files. 

Considering http4s has support for webjars, and the swagger-ui is distributed in its own webjar, we should be possible to provide an example that uses said webjar.

zarthross:
Thanks for opening an issue.  I agree we should at least provide an example using a webjar... what would probably be even better is a module that just hosts the swagger-ui from a webjar to reduce boilerplate for users.

PR #385, opened 2020-02-07
swagger-ui module

RafalSumislawski:
Closes: #366


swagger-ui/src/main/scala/com/http4s/rho/swagger/ui/SwaggerUiRoutes.scala
```diff
@@ -0,0 +1,111 @@
...
+
+object SwaggerUiRoutes {
+
+  def apply[F[+ _] : Sync : ContextShift](swaggerUrl: String, blocker: Blocker): SwaggerUiRoutes[F] = {
+    val swaggerUiResourcesPath = s"/META-INF/resources/webjars/swagger-ui/${BuildInfo.swaggerUiVersion}/"
+    val indexHtml = defaultIndexHtml(swaggerUrl)
+    new SwaggerUiRoutes[F](swaggerUiResourcesPath, indexHtml, blocker)
+  }
+
+  case class Html(html: String)
```
zarthross: I generally like like the way a good type alias can impact the type signature and make it more readable, but is this buying us anything here?  Generally we will only define the swagger-ui once, we aren't checking that the indexHtml is valid like what might be done with an opaque alias...

RafalSumislawski:
The pulling out of SwaggerMetadata didn't affect directly the `org.http4s.rho.swagger.models.Swagger`. It still contains all the `info`, `host`, for which there's no good Monoid definition. 

The way `SwaggerModelsBuilder` is implemented shows, that there should be some Monoid hiding in the `Swagger`.  I think the pair of paths and definitons, let's call it `PathsAndDefinitions` could form a well defined monoid. `SwaggerModelsBuilder` could operate on `PathsAndDefinitions`, and later we could merge the `PathsAndDefinitions` with `SwaggerMetadata ` to get a `Swagger`.

The old URL to swagger-ui should work and redirect to /swagger-ui/index.html. It doesn't, and it has something to do with prepending `swaggerUiTypedPath` in createRhoMiddleware. I'll debug it.

swagger-ui/src/main/scala/com/http4s/rho/swagger/ui/SwaggerUiRoutes.scala
RafalSumislawski: The sole purpose of the `Html` is to create a new type for the implicit search of `EntityEncoder`. The indexHtml is a `String`, but it needs a custom encoder (with html content type). 

Of course there are different ways of achieving that goal, for example:
```
  private val htmlEncoder: EntityEncoder[F, String] =
    EntityEncoder.stringEncoder[F]
      .withContentType(`Content-Type`(MediaType.text.html).withCharset(org.http4s.Charset.`UTF-8`))

  GET / "index.html" |>> { () =>
    Ok(indexHtml)(implicitly, htmlEncoder)
  }
```
would be an alternative way of doing it.

zarthross: That makes sense.  I think I would bias this explicit approach as to not burden users with the wrapping... especially for something that is used only once like this.

RafalSumislawski: Ok for me.

RafalSumislawski:
Hello @zarthross 
Is there something more I should do before we can merge this one?

PR #417, opened 2020-05-10
Improve HListToFunc errors

RafalSumislawski:
In my experience one of the hardest (especially for new users) problems when working with rho are the `could not find implicit value for parameter hltf` compile-time errors. IMO the reason for that is that:
1. The error message speaks about things that the developer isn't dealing directly with. (The `hltf` is a complex implementation detail of a seemingly simple DSL operator `|>>`.)
2. The same error message is reported for multiple classes of errors:
- number of function parameters not matching the route declaration
- types of function parameters not matching the route declaration
- returning value of a type which has no ResultMatcher

I would like to improve that experience by providing error messages which would point the developer in the direction of the root cause of the problem.

### Example 1:
Code:
```
  GET / pathVar[Int] |>> { (i: Int, j: Int) => "Hello World" }
```

Old error message:
```
could not find implicit value for parameter hltf: org.http4s.rho.bits.HListToFunc[F,Int :: shapeless.HNil,(Int, Int) => String]
```

New error message:
```
Expecting a function with parameter types matching (Int :: shapeless.HNil) or (Request[F] :: Int :: shapeless.HNil). Got ((Int, Int) => String).
```

### Example 2:
Code:
```
  GET / pathVar[Int] |>> { (i: Int) => Foo("a", "b")}
```

Old error message:
```
could not find implicit value for parameter hltf: org.http4s.rho.bits.HListToFunc[F,Int :: shapeless.HNil,Int => com.http4s.rho.swagger.demo.MyRoutes.Foo]
```

New error message:
```
No ResultMatcher instance found for the return type of (Int => com.http4s.rho.swagger.demo.MyRoutes.Foo).
```


RafalSumislawski:
The implementation is pretty straightforward, but it does a lot of redundant shapeless things, so I'm a bit worried about the impact this will have on compile times. I have some ideas on how to implement it differently (maybe faster, maybe slower), but I first need to find some reasonable way of benchmarking the compilation.

zarthross:
Thanks for taking on this effort!  The error messages are also something that has caused issues in my usage as well.  I'll take a closer look at this PR when I can, but i certainly don't want to increase the already slow compiles much more than we have to.

RafalSumislawski:
So, I did some benchmarks (https://github.com/RafalSumislawski/rho/tree/compilation-benchmarks/benchmark/src/main/scala/com/http4s/rho/swagger/demo) of the compilation of the rho DSL. They are very simple (read imprecise), but they should be enough to give us a rough understanding of what is expensive to compile and what is not.

The benchmarks are files with the same definition repeated 100 times. The definitions are in batches of 10, so that the whole bytecode doesn't get emitted in one method. That would exceed the method size limit. I start sbt in interactive mode, warm it up with a few compilations, and then repeatedly add/remove an empty line to the benchmark file and recompile it with `rho-benchmark/compile`. If I keep getting the same total time printed by sbt, that is my result.

In all benchmarks I keep `cats.implicits._` in scope. I think that would be the common case. And it is important because it slows down the compilation significantly.

Short initial analysis

Generally, scalac spends the most of time searching for implicits. That is especially true fo projects full of shapless, cats and types with 60 type params (yes, I'm looking at you rho).
The main concern is how long it takes to compile a typical route definition. The cost of compiling a route definition will comprise mainly of:
1. building an HList representing the input (everything before `|>>`)
2. finding an HListToFunction (the `|>>`)
   2a. finding a FnToProduct and a Reverse
   2b. finding a ResultMatcher

To keep is short, I will focus on the few benchmarks which let us decompose the cost:
`Benchmark1_TypedPath` - just 1.
`Benchmark7_ResultMatcherFOk` - 2b in case of a Ok("")
`Benchmark2_TypedPathAndHandler` - 1 + 2a + 2b (still Ok with a string)

Results with current version of rho are:
`Benchmark1_TypedPath` - 3s
`Benchmark7_ResultMatcherFOk` - 7s
`Benchmark2_TypedPathAndHandler` - 10s

Apparently 2a is small compared to 1 and 2b 

My initial implementation (https://github.com/http4s/rho/pull/417/commits/a7dc68f24239f4067d01bfaed26f36e7ca8b3e66) doubled both 2a (FuncParamsMatch) and 2b (FuncResultMatch). And with no surprise, it took 17 seconds in the `Benchmark2_TypedPathAndHandler`.  Doubling 2b is what we definitely need to avoid. Fortunately it's easy (https://github.com/http4s/rho/pull/417/commits/3a65904e8748da0b375c1f6c7c58c619de9c13fd). And with that change, we're back at 10s in `Benchmark2_TypedPathAndHandler`. 

So the current solution is doubles 2a. It surely costs a bit, but it looks like it's insignificant compared to the costs of 1 and 2b. IMO in terms of impact on compile times, this PR is now good to go.

After doing these simple tests, I did actually pull out the big guns against the 2b, but that will a topic of a separate PR.

RafalSumislawski:
BTW I've changed the message for the "Example 2" to:
```
No HListToFunc instance found. Most likely no ResultMatcher instance is available for the return type of (Int => com.http4s.rho.swagger.demo.MyRoutes.Foo). This can be caused by the lack of a matching EntityEncoder.
```

core/src/main/scala/org/http4s/rho/bits/FuncParamsMatch.scala
```diff
@@ -0,0 +1,22 @@
+package org.http4s
+package rho.bits
+
+import shapeless.HList
+
+import scala.annotation.implicitNotFound
+
+@implicitNotFound("Expecting a function with parameter types matching (${T}) or (Request[F] :: ${T}). Got (${FU}).")
```
zarthross, line 8:
```suggestion
@implicitNotFound(
"""Could not bind route to action;
  Expecting a function with parameter types matching:
         (${T}) 
  or     (Request[F] :: ${T})
  found: ${FU}""")
```

Which will yield:
```
[E] [E2] examples/src/main/scala/com/http4s/rho/swagger/demo/MyRoutes.scala:22:22
[E]      Could not bind route to action;
[E]        Expecting a function with parameter types matching:
[E]               (Int :: shapeless.HNil) 
[E]        or     (Request[F] :: Int :: shapeless.HNil)
[E]        found: (Int, Int) => String
[E]      L22:   GET / pathVar[Int] |>> { (i: Int, j: Int) => "Hello World" }
```

I think the extra new lines make it a bit easier to read, and also aligning the types will help somewhat.

core/src/main/scala/org/http4s/rho/bits/HListToFunc.scala
```diff
@@ -5,11 +5,14 @@ import cats.Monad
...
 import shapeless.HList
 
+import scala.annotation.implicitNotFound
+
 /** Converter of an value of type F to the HList of type T
   *
  * @tparam T HList type of the incoming values
  * @tparam U type of element onto which T will be mapped
  */
+@implicitNotFound("No HListToFunc instance found. Most likely no ResultMatcher instance is available for the return type of (${U}). This can be caused by a lack of a matching EntityEncoder.")
```
zarthross, line 15:
```suggestion
@implicitNotFound(
"""Could not bind route to action;
No HListToFunc instance found. 
  Most likely no ResultMatcher instance is available for the return type of 
    (${U}). 
  This can be caused by a lack of a matching EntityEncoder.""")
```

Yields:
```
[E] [E3] examples/src/main/scala/com/http4s/rho/swagger/demo/MyRoutes.scala:24:22
[E]      Could not bind route to action;
[E]      No HListToFunc instance found. 
[E]        Most likely no ResultMatcher instance is available for the return type of 
[E]          (Int => com.http4s.rho.swagger.demo.MyRoutes.Foo). 
[E]        This can be caused by a lack of a matching EntityEncoder.
[E]      L24:   GET / pathVar[Int] |>> { (i: Int) => Foo("a", "b")} 
[E]             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

PR #418, opened 2020-05-10
Add Path.flatMapOperations to make editing Path more convenient

RafalSumislawski:


swagger/src/main/scala/org/http4s/rho/swagger/models.scala
```diff
@@ -254,6 +254,17 @@ object models {
       vendorExtensions.foreach { case (key, value) => p.setVendorExtension(key, value) }
       p
     }
+
+    def flatMapOperations(f: Operation => Option[Operation]): Path =
```
zarthross, line 258: Not opposed to putting in this code, but naming wise, I think flat map is not appropriate.  This is probably more akin to a `collect` ?

RafalSumislawski: `collestSome` maybe? The signature is more similar to [collestFirstSome](https://github.com/typelevel/cats/blob/a0752346dd0eb2aca54c5a2e153340694d33bada/core/src/main/scala/cats/Foldable.scala#L265) or [collectFoldSome](https://github.com/typelevel/cats/blob/a0752346dd0eb2aca54c5a2e153340694d33bada/core/src/main/scala/cats/Foldable.scala#L333) than to `collect/collectFirst/collectFold` which all take a PartialFunction as an argument.

zarthross: Yeah, `collectSomeOperations` ?  It's wordy but I think captures better than flatMap here.

zarthross (approved):

zarthross:

PR #453, opened 2020-10-03
Make names of generic types customisable

### LINKED ISSUE #416: Feature Idea: Customisable generic type names
Currently when we have a generic type `Foo[T]` and we parametrize it with `Bar` (`Foo[Bar]`), when what we get in the swagger file is a `Foo«Bar»`. This name contains two characters outside of the ASCII, which causes problems with some tools (https://editor.swagger.io/ for example). I think it would be nice to make the strategy of composing names for generic types customizable. For example, one may want to render `Foo[Bar]` as `Foo<Bar>` or `FooOfBar` or `Foo_Bar`.

RafalSumislawski:
The default behaviour remains unchanged. The way names are generated can be customised by providing a custom `ShowType` instance. This customisability is not limited to generic types, so if someone wishes to let's say, convert CamelCase class names into kebab-case type names in swagger that's also possible.  

Internally the `ShowType` is propagated as an implicit. On the API surface it's handled analogically to `SwaggerFormats` (non-implicit param with a default value).

Fixes #416


zarthross (approved):

zarthross:

Issue #448, opened 2020-09-10
Simple or-path produces incorrect tags


mcenkar:
Hi,

When using simple or-path:

`GET / ("foo" || "bar") |>> { () => "Hello!" }`

following `swagger.json` is produced:

```
{
  "paths": {
    "/foo": {
      "get": {
        "tags": [
          "foo",
          "bar"
        ],
        "operationId": "getFoo",
        "produces": [
          "text/plain"
        ]
      }
    },
    "/bar": {
      "get": {
        "tags": [
          "foo",
          "bar"
        ],
        "operationId": "getBar",
        "produces": [
          "text/plain"
        ]
      }
    }
  }
}
```

`/foo` path gets `bar` tag and `/bar` path gets `foo` tag which leads to unnecessary entries in swagger ui:

![image](https://user-images.githubusercontent.com/11230749/92688620-f5c36780-f33d-11ea-8ab5-ae46406b86d0.png)

I would expect single tags on both routes.

Also if you specify tag yourself:

`  "mytag" @@ GET / ("foo" || "bar") |>> { () => "Hello!" }`

it ends up there twice:

```
{
  "/foo": {
    "get": {
      "tags": [
        "mytag",
        "mytag"
      ],
      "operationId": "getFoo",
      "produces": [
        "text/plain"
      ]
    }
  }
}
```

zarthross:
Fixed in #454

Issue #165, opened 2017-01-21
Swagger doesn't work well with `||` operators


bryce-anderson:
This can be observed via broken swagger docs for routes that have such definitions. This brings me to the important question: does the `||` logic serve a real purpose or is it just a cool trick? I'm pretty pleased to be able to do it, but most of the examples I can think of are highly contrived. That would be fine except it brings significant additional complexity, thus this bug.

Do any of the rho users actually use the `||` path logic? I know one of our most heavily invested users didn't even know it existed.

PR #454, opened 2020-10-03
Fix generation of swagger for alternative routes (routes with || )

RafalSumislawski:
This is a fix for #448. 

There are already tests ("find a simple or-path" and "find a capture or-path") which exercise the scenario leading to this issue, but due to the way the assertions are formulated the problem is not detected. Redefinition of these tests is also part of this PR.

The issue actually has impact not only on tags, but generally on every difference between the alternatives defined with `||` . The root cause of the issue is that `SwaggerModelsBuilder.mkOperation` works on a `RhoRoute`, while a single `RhoRoute` may represent more than one operation in swagger model. Information belonging to these alternative operations gets mixed up. The solution is to split single `RhoRoute` into multiple routes, each representing one of the alternatives (I named them `LinearRoute`s) and then building the operation basing on these. This is more or less what `collectPaths -> mkPathStrs -> linearizeStack` call chain already does, but the distinction between alternative paths is somewhat lost instead of being propagated to `mkOperation`. 

The fix itself (see first commit) is relatively simple, but it requires modifying the signature of `mkOperation` and that method is used 28 times in tests. So the real challenge here was to implement the fix in a way which would not make the tests overly complicated/verbose. Hence the sizeable refactor in the second commit.

### DIFF
--- a/swagger/src/main/scala/org/http4s/rho/swagger/SwaggerModelsBuilder.scala
+++ b/swagger/src/main/scala/org/http4s/rho/swagger/SwaggerModelsBuilder.scala
@@ -14,22 +14,21 @@ import scala.collection.immutable.Seq
 import scala.reflect.runtime.universe._
 import scala.util.control.NonFatal
 
-private[swagger] class SwaggerModelsBuilder(formats: SwaggerFormats) {
+private[swagger] class SwaggerModelsBuilder[F[_]](formats: SwaggerFormats)(implicit etag: WeakTypeTag[F[_]]) {
   import models._
 
   private[this] val logger = getLogger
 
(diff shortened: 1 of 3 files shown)


RafalSumislawski:
I think this PR may also fix the #165. At least the swagger-not-working-well part. It obviously doesn't solve the do-we-even-need-this-feature part.

RafalSumislawski:
Hi @zarthross 
Is there anything more I should do about this (and the others) PR before it can be merged? I think the Binary Compatibility has nothing to do with the changes in this PR, but if that's a blocker, I can try doing something about it.

zarthross (approved):

zarthross:
Hey, sorry for the delay... lots of things have come up for me... I'll dedicate some time this week to getting your PRs rolled out into a release.

zarthross:

PR #476, opened 2020-11-24
Relax variance constraint on F in SwaggerUi/SwaggerUiRoutes

RafalSumislawski:
I've noticed that `F` being covariant (`F[+_]`) in `SwaggerUi` and `SwaggerUiRoutes` is not only unnecessary (the covarience is not used in any way), but it causes issues when working with an `F` which is invariant, like for example `ReaderT[IO, A]`. Therefore in this PR I'm proposing to remove the requirment of covariance.


zarthross (approved):

zarthross:

PR #560, opened 2021-06-29
Use MUnit in `rho-swagger` tests

danicheg:
The main goal of this PR is a unification of the test framework in `http4s`'s projects.
Changing the test framework for the `rho-core` module is better for making it in another PR.


swagger/src/test/scala/org/http4s/rho/swagger/TypeBuilderSuite.scala
```diff
@@ -0,0 +1,540 @@
...
+      typeTag[Symbol],
+      typeTag[java.math.BigDecimal],
+      typeTag[java.math.BigInteger]
+    )
+
+    val ms = primitives.foldLeft(Set.empty[Model]) { (s, t) =>
+      TypeBuilder.collectModels(t.tpe, s, DefaultSwaggerFormats, typeOf[IO[_]])
+    }
+
+    assert(ms.isEmpty)
```
zarthross, line 115: Could we refactor this (and maybe other `.isEmpty` tests so that they `assertEquals(ms, List.empty)` (roughly speaking).  This way when the assertion fails we can see what the values are that were not supposed to be there. 
 I would also be open to adding some additional `clue`s.

zarthross (commented):
Thanks for submitting a PR, this looks great!  Just had two minor comments, but nothing major.

If you feel like working on the other tests too, I would certainly be grateful!

swagger/src/test/scala/org/http4s/rho/swagger/TypeBuilderSuite.scala
danicheg: Good point, I've fixed that.

danicheg:
@zarthross Thanks for the review. It would be great if you'll take a look once more.

zarthross (approved):

Issue #573, opened 2021-07-27
Non-class Scala types break TypeBuilder


jarrodcodes:
Recently I switched to Rho for Swagger doc generation. My models include Refined types that had exceptions while building their Swagger models. The main reason this was breaking was because the Refined types were not classes and this case was not handled. I propose adding a check for this and handling it with the provided fallback method.

With this change: 
![image](https://user-images.githubusercontent.com/1577741/127079346-c95d585e-332b-4859-bae1-783087827c42.png)

Without this change: 
![image](https://user-images.githubusercontent.com/1577741/127079422-ac447b7d-b8c7-47c7-9967-a69d72c74626.png)

PR #574, opened 2021-07-27
#573 Handle nonclass Scala types and add tests

jarrodcodes:
This fixes https://github.com/http4s/rho/issues/573. All tests + 2 new ones are passing.

### DIFF
--- a/swagger/src/main/scala/org/http4s/rho/swagger/TypeBuilder.scala
+++ b/swagger/src/main/scala/org/http4s/rho/swagger/TypeBuilder.scala
@@ -19,16 +19,16 @@ object TypeBuilder {
   private[this] val logger = getLogger
 
   def collectModels(t: Type, alreadyKnown: Set[Model], sfs: SwaggerFormats, et: Type)(implicit
-      st: ShowType): Set[Model] =
+                                                                                      st: ShowType): Set[Model] =
     try collectModels(t.dealias, alreadyKnown, ListSet.empty, sfs, et)
     catch { case NonFatal(_) => Set.empty }
 
   private def collectModels(
-      t: Type,
-      alreadyKnown: Set[Model],
-      known: TypeSet,
-      sfs: SwaggerFormats,
-      et: Type)(implicit st: ShowType): Set[Model] = {
+                             t: Type,
+                             alreadyKnown: Set[Model],
+                             known: TypeSet,
+                             sfs: SwaggerFormats,
+                             et: Type)(implicit st: ShowType): Set[Model] = {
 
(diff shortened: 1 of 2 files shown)


swagger/src/main/scala/org/http4s/rho/swagger/TypeBuilder.scala
```diff
@@ -210,27 +210,34 @@ object TypeBuilder {
...
-          DataType.fromType(tpe) match {
-            case DataType.ValueDataType(name, format, qName) =>
-              AbstractProperty(`type` = name, description = qName, format = format)
-            case DataType.ComplexDataType(name, qName) =>
-              AbstractProperty(`type` = name, description = qName)
-            case DataType.ContainerDataType(name, _, _) =>
-              AbstractProperty(`type` = name)
-            case DataType.EnumDataType(enums) =>
-              StringProperty(enums = enums)
+        else if (tpe.isAnyVal && !tpe.isPrimitive && ptSym.isClass) {
```
jarrodcodes, line 213: My Refined type was not a class, therefore `ptSym.asClass` would break.

swagger/src/main/scala/org/http4s/rho/swagger/TypeBuilder.scala
```diff
@@ -312,21 +318,26 @@ object TypeBuilder {
         if (t.typeArgs.nonEmpty) GenArray(fromType(t.typeArgs(1)))
         else GenArray()
       } else if (klass <:< typeOf[AnyVal]) {
-        fromType(
```
jarrodcodes, line 315: When testing my fix I also discovered that `AnyValClass` breaks here. Therefore we can also use the fallback in this case as well.

swagger/src/main/scala/org/http4s/rho/swagger/TypeBuilder.scala
```diff
@@ -312,21 +318,26 @@ object TypeBuilder {
...
       } else if (isObjectEnum(klass.typeSymbol)) {
         EnumDataType(klass.typeSymbol.asClass.knownDirectSubclasses.map(_.name.toString))
       } else {
-        val stt = if (t.isOption) t.typeArgs.head else t
-        ComplexDataType("string", qualifiedName = Option(stt.fullName))
+        fallBackDataTypeFromName(t)
       }
     }
 
+    private def fallBackDataTypeFromName(t: Type)(implicit st: ShowType): DataType = {
```
jarrodcodes, line 336: I propose this name because this Swagger dataType is clearly a last result String that only displays the class name, which is far better than nothing, but should be plainly labeled as a last resort.

zarthross (approved):
Looks good to me!

PR #598, opened 2021-09-22
Fixes class cast exception on capture tail being

vaslabs:
Hello,
We've recently updated to 0.22 and we caught this class cast exception error where the returned list of remaining path is a List of segments but it gets passed into a List of Strings via shapeless, causing a class cast exception if you actually map over the list or do anything that would create a String field iload to be accessed in bytecode.

I believe this fixes it.

I haven't checked recent versions if it's fixed since this is the one we are using for now.

I've left a println in the examples `println(l.map(_.getClass))` which if you add in the current code base will throw the class cast exception if you want to verify.

Thank you for this project and please let me know your thoughts.


vaslabs:
edit: I removed the println to make it mergeable

core/src/test/scala/org/http4s/rho/RhoRoutesSuite.scala
```diff
@@ -63,7 +63,7 @@ class RhoRoutesSuite extends CatsEffectSuite with RequestRunner {
         Ok("twoparams2_" + foo + bar.getOrElse("cat"))
     }
 
-    GET / "variadic" / * |>> { tail: List[String] => Ok("route8_" + tail.mkString("/")) }
+    GET / "variadic" / * |>> { tail: List[Segment] => Ok("route8_" + tail.map(identity).mkString("/")) }
```
FrancescoSerra, line 66: I'm a bit confused as to why you need to `.map(identity)` here?

FrancescoSerra (commented):
Thanks for the contribution, and sorry it took so long to come back to you - I'm trying to catch up with everything in this repo! :-)

core/src/test/scala/org/http4s/rho/RhoRoutesSuite.scala
vaslabs: it's just a simple way to force the use of the tail type, so if it doesn't match it will throw a runtime error

vaslabs: > Thanks for the contribution, and sorry it took so long to come back to you - I'm trying to catch up with everything in this repo! :-)

no worries, we moved to 0.23 since then and the same issue exists there too

FrancescoSerra: Right - I think we should target 0.23 with this, as I'm thinking of releasing soon against series 0.23. Thoughts?

vaslabs: I agree, let me change the target branch, see if there are any conflicts

zarthross: I think we should do this PR for Series/0.22, Series/0.23, and main.

zarthross: I think this is more obvious and performs the same duty.  Lets change it to this.

```suggestion
    GET / "variadic" / * |>> { tail: List[Segment] => Ok("route8_" + tail.map(_.encoded).mkString("/")) }
```

zarthross (approved):
Approved, but lets change that test before we merge.  Thanks for the work @vaslabs !

core/src/test/scala/org/http4s/rho/RhoRoutesSuite.scala
vaslabs: yeah, this was just to demonstrate that picking the tail type by any means was breaking. when I was writing the test, .encoded wouldn't make much sense for proving the issue.
but .encoded going forwards will complain if the type changes again so I'll change to that

swagger-ui/src/main/scala/com/http4s/rho/swagger/ui/SwaggerUiRoutes.scala
```diff
@@ -30,7 +31,7 @@ class SwaggerUiRoutes[F[_]: Sync: ContextShift](
     Ok(indexHtml)(implicitly, htmlEncoder)
   }
 
-  GET / swaggerUiPath / CaptureTail |>> { (req: Request[F], path: List[String]) =>
+  GET / swaggerUiPath / CaptureTail |>> { (req: Request[F], path: List[Segment]) =>
     fetchResource(swaggerUiResourcesPath + path.mkString("/", "/", ""), req)
```
vaslabs, line 35: I think this can be done better now, like `swaggerUiResourcesPath / path`

FrancescoSerra:
@vaslabs I can't merge your PR because some of your commits were not signed. Would you mind squashing the intermediate commits and then push back up? This should make GH happy

Issue #606, opened 2021-10-07
Logger Options in RhoRoutes


blmeadows:
Is having this final logger a requirement/strong preference for the RhoRoutes? 

https://github.com/http4s/rho/blob/9f078b2277cdf9ef21694c3347b37a3082d520a9/core/src/main/scala/org/http4s/rho/RhoRoutes.scala#L33

PR #607, opened 2021-10-12
Remove prescriptive RhoRoutes logger

blmeadows:
update for https://github.com/http4s/rho/issues/606


zarthross (approved):

zarthross:
Thanks for the PR!  This is binary breaking so it'll have to wait for the next bin-incompat release.

FrancescoSerra:
Considering that now `main` target the `1.0.0-M*` series, which has no bin-compat guarantees, I think this should be fine to merge. @zarthross thoughts?

PR #668, opened 2022-03-16
Use Scala Collection Compat

etspaceman:
Currently, providing the CollectionConverters object manually in the `scala.jdk` package as a part of Rho causes assembly issues, as projects with dependencies on Rho and the Scala Collection Compat will require conflicts to be resolved. 

From what I understand, this was originally done to avoid bincompat issues during updates, as the collection compat library broke some rules early on. This has gotten a lot more stable and is pretty widely used now in the Scala ecosystem, so it seems much safer to include this library.

Another option could be to repackage the collection converters under a new namespace. I could work on that if that is preferred.

### DIFF
--- a/core/src/main/scala-2.12-/jdk/CollectionConverters.scala
+++ b/core/src/main/scala-2.12-/jdk/CollectionConverters.scala
@@ -1,5 +1,0 @@
-package scala.jdk
-
-import scala.collection.convert.{DecorateAsJava, DecorateAsScala}
-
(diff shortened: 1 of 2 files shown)


etspaceman:
Looks like this fails mima so we would have to be OK breaking binary compat for this kind of change. It seems pretty darn harmless though, considering the file that is being removed is replaced by a transitive dependency in the same package.

PR #714, opened 2022-07-12
Main dependencies update

MarmaladeSky:
This PR includes:
- update http4s to 1.0.0-M34 and common dependencies to the same versions as http4s has
- correction of the PathTree due to changes in http4s Path
- update Scala to 2.13.8, remove Scala 2.12 support (it was dropped in http4s since 1.0.0-M31)


zarthross (commented):
Thanks for the PR.  This looks pretty good, but I need to double check the `splitPath` changes.   I'll do this when I'm able to find some time.

FrancescoSerra:
Many thanks for your contribution @MarmaladeSky !

MarmaladeSky:
Thanks, looking forward to the release.

PR #761, opened 2022-09-05
Fix/tail capture segment class cast

vaslabs:
See https://github.com/http4s/rho/pull/598 but for 0.23

### DIFF
--- a/core/src/test/scala/org/http4s/rho/RhoRoutesSuite.scala
+++ b/core/src/test/scala/org/http4s/rho/RhoRoutesSuite.scala
@@ -2,13 +2,13 @@ package org.http4s
 package rho
 
 import java.util.concurrent.atomic.AtomicInteger
-
 import cats.effect.IO
 import fs2.Stream
 import munit.CatsEffectSuite
 import org.http4s.headers.{`Content-Length`, `Content-Type`}
 import org.http4s.rho.io._
 import org.http4s.Uri.Path
+import org.http4s.Uri.Path.Segment
 import org.typelevel.ci.CIString
 
 import scala.collection.compat.immutable.ArraySeq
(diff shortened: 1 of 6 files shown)


zarthross (approved):
Good work!   Normally I'd flag this a bin-incompat (even though mima disagrees), changing from String to Segment could break a bunch of stuff, but since the last release was RC1... I  think we can get away with sneaking this into Series/0.23.

zarthross:
@FrancescoSerra Any thoughts here?   Let's make sure to forward port this into main.

core/src/test/scala/org/http4s/rho/RhoRoutesSuite.scala
```diff
@@ -63,7 +63,9 @@ class RhoRoutesSuite extends CatsEffectSuite with RequestRunner {
         Ok("twoparams2_" + foo + bar.getOrElse("cat"))
     }
 
-    GET / "variadic" / * |>> { tail: List[String] => Ok("route8_" + tail.mkString("/")) }
+    //identity is used to force the type usage of the path tail
+    GET / "variadic" / * |>> { tail: List[Segment] =>
+      Ok("route8_" + tail.map(identity).mkString("/")) }
```
zarthross, line 68:
```suggestion
    //encoded is used to force the type usage of the path tail
    GET / "variadic" / * |>> { tail: List[Segment] =>
      Ok("route8_" + tail.map(_.encoded).mkString("/")) }
```

vaslabs:
> Good work! Normally I'd flag this a bin-incompat (even though mima disagrees), changing from String to Segment could break a bunch of stuff, but since the last release was RC1... I think we can get away with sneaking this into Series/0.23.

I would agree but I thought that the binary incompatibility broke before this change, maybe with some http4s update? something that switched the tail type from string to segment, I didn't go that deep into the investigation. What I mean, this PR doesn't really introduce binary incompatibility

core/src/test/scala/org/http4s/rho/RhoRoutesSuite.scala
FrancescoSerra: @vaslabs are you happy with incorporating @zarthross ' suggestion?

vaslabs: I thought I pushed it already but maybe I only did on the other branch

vaslabs: done

PR #1185, opened 2025-06-08
Publish to Central Portal

rossabaker:
Things I'm not proud of herein:

1. Drops Scala 2.12 support
2. Adds a bunch of dubious `@nowarn` around "The outer reference in this type test cannot be checked at run time." errors
3. Adds a dubious `@nowarn` to an entire suite that uses a `val Some(...)` idiom
4. Gives up building a site


rossabaker:
It looks like I've legitimately broken some tests.

rossabaker:
Going to try again, leaner

